PR #141, opened 2026-03-08
feat: hook-aware converter for cross-platform hook dependencies

### LINKED ISSUE #138: bug: Codex/OpenCode commands reference non-existent context hook output
## Summary

Codex and OpenCode skill/command prompts reference context hook output that doesn't exist on those platforms. The prompts were converted from Claude Code commands and still contain:

> **Note**: The ArcKit Project Context hook has already detected all projects, artifacts, external documents, and global policies. Use that context below — no need to scan directories manually.

But Codex and OpenCode **have no context hook** — so there's no injected context to use. This means external documents in `projects/*/external/` may not be automatically detected.

## Affected Platforms

| Platform | Context Hook | External Doc Detection |
|----------|-------------|----------------------|
| Claude Code | `arckit-context.mjs` (UserPromptSubmit) | Working — lists external files automatically |
| Gemini | `context-inject.py` (BeforeAgent) | Working — lists external files automatically |
| **Codex** | **None** | **Broken** — prompt references missing context |
| **OpenCode** | **None** | **Broken** — prompt references missing context |

## Root Cause

The converter (`scripts/converter.py`) copies the Claude Code command prompts and rewrites paths, but doesn't strip or replace the context hook reference. The note about "context hook has already detected..." is meaningless on platforms without hooks.

## Proposed Fix

Update the converter to replace the context hook reference in Codex/OpenCode output with explicit directory scanning instructions, e.g.:

> **Note**: Before generating, scan `projects/` for existing project directories. For each project, check `external/` for reference documents and list all existing `ARC-*.md` artifacts.

The prompts should also **ask the user** when no external files are found, e.g.:

> "No external documents found in `projects/{project}/external/`. Do you have any reference documents (RFPs, specs, existing designs) to include? Place them in the `external/` folder and re-run, or continue without them."

This ensures the user is always prompted rather than silently generating without key inputs.

## Short-term Workaround

Until the fix is in place, Codex/OpenCode users can explicitly reference their external documents when running commands:

```
$arckit-requirements Use the documents in projects/001-myproject/external/ as input
```

## Affected Commands

All commands that reference external documents or project context — at minimum:
- requirements
- stakeholders
- risk
- sobc
- data-model
- plan

Discovered via #120.

### LINKED ISSUE #139: bug: /arckit.pages command is a no-op on Codex/Gemini/OpenCode — relies on Claude-only hook
## Summary

The `/arckit.pages` command is entirely dependent on the `sync-guides` UserPromptSubmit hook, which only exists in Claude Code. On Codex, Gemini, and OpenCode the hook never fires, so the command does nothing useful.

The converted prompts on all non-Claude platforms still contain:

> The `sync-guides` hook handles ALL I/O before this command runs

> **Do NOT call any tools** — no Read, no Glob, no Write, no Bash. The hook's stats are the single source of truth.

This means the AI is explicitly told to do nothing — making the command a complete no-op.

## Affected Platforms

| Platform | `sync-guides` hook? | Result |
|----------|-------------------|--------|
| Claude Code | Yes (UserPromptSubmit) | Working |
| **Gemini** | No | **No-op** — AI told not to call tools |
| **Codex** | No | **No-op** — same |
| **OpenCode** | No | **No-op** — same |

## Root Cause

The converter (`scripts/converter.py`) copies the Claude command prompt verbatim. For most commands this works fine since the hook just adds context. But `/arckit.pages` is different — the hook does **all the work** (guide sync, template processing, manifest generation, HTML output). The command prompt is just a summary reporter.

## Proposed Fix

Two options:

**Option A: Inline the hook logic into non-Claude prompts.** The converter replaces the "hook did everything" section with explicit instructions to:
1. Copy guides from the extension directory to `docs/guides/`
2. Read `pages-template.html` and process it
3. Scan projects and build manifest
4. Write `docs/index.html`

**Option B: Remove the command from non-Claude platforms.** Since the command is fundamentally hook-dependent, it may not be worth converting. Add it to an exclusion list in the converter.

## Related Issues

- #138 — Codex/OpenCode commands reference non-existent context hook output (same class of bug)
- #122 — Original user report of `/arckit.pages` failing in Cursor

tractorjuice:
## Summary

- Adds `has_context_hook` and `has_sync_guides_hook` flags to `AGENT_CONFIG` so the converter knows which platforms have hooks
- Adds `rewrite_hook_dependencies()` that replaces the context injection hook note with self-scan instructions for 43 commands on hookless platforms (Codex, OpenCode)
- Adds `commands-standalone/` directory with a standalone `pages.md` (pre-hook version with full inline logic) for platforms without the sync-guides hook
- Gemini retains the original hook note since it has `context-inject.py`

Closes #138, closes #139

## Design

See `docs/plans/2026-03-08-hook-aware-converter-design.md`

## Test Plan

- [x] Converter runs without errors (228 files generated)
- [x] 43 Codex/OpenCode commands have self-scan replacement
- [x] 0 Codex/OpenCode commands still reference the hook
- [x] 43 Gemini commands retain the original hook note
- [x] Codex/OpenCode pages command uses standalone version (no "Handled by Hook")
- [x] Codex skills also get the replacement (43 confirmed)

### DIFF
--- a/arckit-claude/commands-standalone/pages.md
+++ b/arckit-claude/commands-standalone/pages.md
@@ -1,0 +1,568 @@
+---
+description: Generate documentation site with governance dashboard, document viewer, and Mermaid diagram support
+allowed-tools: Read, Write, Glob, Grep
+argument-hint: "<project ID or 'all', e.g. '001', 'all'>"
+---
+
+# ArcKit: Documentation Site Generator
+
+You are an expert web developer helping generate a documentation site that displays all ArcKit project documents with full Mermaid diagram rendering support.
+
+## What is the Pages Generator?
+
+The Pages Generator creates a `docs/index.html` file that:
+
+- **Dashboard** with KPI cards, donut charts, coverage bars, and governance checklist
+- **Displays** all ArcKit artifacts in a navigable web interface
+- **Renders** Mermaid diagrams inline
+- **Organizes** documents by project with sidebar navigation
+- **Follows** GOV.UK Design System styling
+- **Works** with any static hosting provider (GitHub Pages, Netlify, Vercel, S3, etc.)
+
+## Your Task
+
+**User Request**: $ARGUMENTS
+
+Generate a documentation site for this ArcKit repository.
+
+## Step 0: Determine Repository Info
+
+Determine the repository name and URL:
+
+1. Read the `.git/config` file and find the `[remote "origin"]` section to get the remote URL
+2. Extract the repo name and owner from the URL (e.g. `https://github.com/owner/repo-name` → repo name is `repo-name`, owner is `owner`)
+3. If `.git/config` doesn't exist or has no remote, use the current directory name as the repo name
+
+## Step 1: Discover Repository Structure
+
+Use **Glob** and **Read** tools to scan the repository. Do NOT use `ls`, `find`, `for` loops, `head`, `grep`, `sed`, or any Bash commands for file discovery.
+
+### 1.1 Guides (Command Documentation)
+
+**Guide sync and title extraction are handled automatically by the `sync-guides` hook** which runs before this command executes. The hook copies all guide `.md` files from the plugin to `docs/guides/` and extracts the first `#` heading from each file — zero tool round-trips.
+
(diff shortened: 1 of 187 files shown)


tractorjuice:
Merged into daily branch (PR #148).

PR #331, opened 2026-04-20
Add Roocode

arnaudgelas:


Drago-cod:
Hi @arnaugleas, thanks for the contribution

I’ve gone through the PR and the additions around Roo Code distribution look really useful, especially aligning the bundle with other distributions and adding scaffolding support.

A few things to note before merging:

- There are **merge conflicts** that need to be resolved.
- The workflow is still **awaiting approval**, so that will need to be cleared.
- It would help to include a brief **PR description** explaining the purpose and impact of these changes for better context.

Once those are addressed, this should be good to move forward. Great work overall!

tractorjuice:
Thanks for the contribution @arnaudgelas — closing this without merge.

Adding a new distribution format (Roocode would become the 8th alongside CLI / Claude Code plugin / Codex / Gemini / OpenCode / Copilot / Paperclip) is a long-term maintenance commitment: every command added to the plugin source has to fan out via `scripts/converter.py`, every extension version has to be bumped via `scripts/bump-version.sh`, and a separate publish path has to be wired into `scripts/push-extensions.sh`. We'd want to scope that against:

1. **Demand signal** — is there an open issue from Roocode users asking for this? I can't see one
2. **Format spec stability** — Roocode's `.roo/rules-*/01-instructions.md` directory pattern is still evolving upstream
3. **PR description** — the PR body is empty, so the integration approach (rules-only? full skill mapping? auto-discovery story?) isn't documented

If you'd like to revisit, please open an issue first describing the use case, the maintenance contract you're proposing, and the upstream Roocode commitment to the directory layout. We can then decide whether the converter gains a new target or whether this should live as a community overlay (similar to the EU/FR overlays in `arckit-claude/commands-community/`).

PR #336, opened 2026-04-20
feat: Wire userConfig into plugin infrastructure

Konuktor:
PR Title
fix: keep templates static placeholders for classification and owner

PR Description
This PR cleans up the userConfig follow-up branch to match plugin substitution rules and remove merge-risk regressions.

What was fixed
Reverted template-level dynamic placeholders across all template packs:
${user_config.default_classification} / ${default_classification} → [CLASSIFICATION]
${user_config.organisation_name} / ${organisation_name} → [OWNER_NAME_AND_ROLE]
Removed template fallback suffixes introduced by the branch so templates remain static.
Kept runtime substitution architecture intact:
scripts/converter.py logic preserved
arckit-claude/hooks/governance-scan.mjs preserved
command/agent user_config guidance preserved
Why
Template files are static scaffolds and do not get runtime user_config substitution. Dynamic placeholders there leaked into generated artifacts and conflicted with plugin behavior.

Validation
No ${user_config.*} remains in templates.
Default classification guidance in commands remains consistent (OFFICIAL for UK Gov, PUBLIC for Generic when unset).
Plugin manifest version remains 4.9.0.


tractorjuice (changes requested):
## Review

Thank you for the thoughtful wiring work — the converter changes and the governance-scan hook additions are solid. However, this PR has one **critical merge blocker** (a stale branch that would revert recent work) and one **architectural concern** about the template-level placeholder approach that I think needs a design discussion before the 60+ template changes land.

Recommending changes requested.

---

### Blocker 1 — Branch is stale and would revert recent work

The branch forked before v4.9.0 and has not been rebased. The diff against `main` includes **unintentional reverts**:

| File | Regression |
|------|------------|
| `pyproject.toml` | Version goes `4.9.0` → `4.8.0` |
| `docs/README.md` | Version footer goes `4.9.0` → `4.8.0` |
| `docs/getting-started.html` | Removes the `--sparse` install tip added in #343 |
| `tests/mermaid-wardley/convert.mjs` | **Deleted entirely** (-442 lines) — this is from PRs #339, #340, #341, #344 |
| `tests/mermaid-wardley/test-fidelity.mjs` | **Deleted entirely** (-457 lines) |
| `tests/mermaid-wardley/test-real-maps.mjs` | Large rewrite that reverts recent fixes |
| `tests/mermaid-wardley/package*.json` | Lock file reverted |

The PR description says "540 files" but the actual diff is **583 files**; the extra 43 are these unintentional reverts.

**Required**: rebase onto current `main` and verify the `git diff` contains only the intended changes.

---

### Blocker 2 — Template-level `${user_config.*}` substitution likely won't work

Per the [Claude Code plugin docs](https://code.claude.com/docs/en/plugins-reference.md#user-configuration), `${user_config.KEY}` is substituted in:

- MCP/LSP server configs
- Hook commands, monitor commands
- **Skill and agent content** (for non-sensitive values)

Templates (`.arckit/templates/*.md`) are **none of these** — they're arbitrary filesystem files read by commands via the `Read` tool at runtime. `Read` returns raw content; no substitution happens.

That means every one of the 60+ templates in this PR now contains a literal string like:

```
| Classification | ${user_config.default_classification} (fallback: [PUBLIC / OFFICIAL / OFFICIAL-SENSITIVE / SECRET]) |
```

When Claude reads the template and copies it into the generated artifact, it will paste that literal string verbatim rather than the configured value. Even if Claude's inference catches the intent sometimes, the compound `${placeholder} (fallback: [...])` pattern is ambiguous enough that some outputs will be wrong.

**The correct architecture** (already used by the command changes in this PR, e.g. `requirements.md` line 195):

- Templates keep the existing `[CLASSIFICATION]`, `[OWNER_NAME_AND_ROLE]` bracket placeholders as-is.
- Command bodies reference `${user_config.*}` when instructing Claude how to fill them, because command bodies **are** substituted. The PR already does this correctly for `requirements.md`, `sobc.md`, and others.

I think the 60+ template changes should be reverted, and the work of surfacing user_config defaults should live entirely in the command instructions (which the PR already does).

**If I've misunderstood the plugin substitution scope, please point me at the docs** — happy to re-review if templates do get substituted somewhere I missed.

---

### What's solid (keep this work)

- **`scripts/converter.py`** — the `rewrite_user_config_placeholders` helper, the in-tree rewrite for copied extensions, and the per-target application look correct. Env-var fallback (`${KEY}`) for non-Claude targets is the right strategy.
- **`arckit-claude/hooks/governance-scan.mjs`** — verified correct:
  - `CLAUDE_PLUGIN_OPTION_*` uppercased naming matches the documented plugin subprocess env var convention
  - All three fields (`organisation_name`, `default_classification`, `governance_framework`) are non-sensitive in `plugin.json`, so they will be exported
  - "Plugin Default Alignment Gaps" section design is genuinely useful — flagging classification/owner mismatches and Generic+UK-Gov artifact collisions
- **Command body updates** for `requirements.md`, `sobc.md`, `tcop.md`, `atrs.md`, `service-assessment.md` — these correctly reference `${user_config.*}` in the instruction text (which IS substituted). This is the right pattern; keep these and remove the template changes.
- **Agent updates** for `arckit-research.md` and `arckit-datascout.md` — same pattern, correct.

---

### Minor nits (for when you rebase)

- **PR description file count**: says "540 files generated correctly" — actual diff shows 583 changed files once the unintended reverts are removed, the real count will drop substantially.
- **Command updates have inconsistent defaults**: `requirements.md` says *"OFFICIAL for UK Gov, PUBLIC for Generic"* while `sobc.md` historically said *"OFFICIAL for UK Gov, INTERNAL for private sector"* and this PR changes it to *"OFFICIAL for UK Gov and PUBLIC for Generic"* — worth one pass to make the per-mode defaults consistent across all updated commands.
- **Worth adding a short section to `docs/guides/` (or `CLAUDE.md`)** explaining the contract: "user_config values are substituted in MCP configs, hook commands, and command/agent bodies — but NOT in template files read at runtime." This would save the next contributor from the same trap.

---

### Proposed path forward

1. Rebase onto `main` to drop the version/test/doc reverts (blocker 1).
2. Revert the 60+ template edits; keep the command-body and agent-body edits which already do the right thing (blocker 2), OR provide evidence that template-level substitution does work in the Claude runtime.
3. Keep the converter and governance-scan.mjs work as-is — both are good.
4. Add a short note to `CLAUDE.md` documenting the user_config substitution contract.

Happy to re-review once the rebase is done.

tractorjuice:
Closing this in its current shape — but the work is largely good and we'd like a follow-up PR with a smaller scope. Thanks @ErbolTakhirov, sticking with you on this if you'd like to land a v2.

## Why we can't merge as-is

### 1. Template-level `\${user_config.X}` substitution doesn't actually run

The HEAD commit (`0d9677db`) puts `\${user_config.default_classification} (fallback: [...])` into 83 of 84 plugin templates. Per [CLAUDE.md's userConfig section](../blob/main/CLAUDE.md), `\${user_config.X}` only substitutes in two places:

- `.mcp.json` (manifest)
- Command and agent **prompt bodies**

Templates are read at runtime via the Read tool — Claude Code doesn't substitute placeholders in arbitrary file contents. So in Claude Code, the literal string `\${user_config.default_classification}` ends up in the rendered artifact (the model *might* infer to substitute, but the behaviour isn't guaranteed). For non-Claude targets (Codex / Gemini / OpenCode / Copilot / Paperclip), the converter rewrites these to bare `\${default_classification}` — but again, **nothing substitutes env vars in markdown files** at those targets either. After the rewrite, 640 extension-template files contain literal `\${default_classification}` strings.

The earlier commit on this branch (`c9c80462` "fix: keep templates static placeholders") had the right pattern: keep templates with `[CLASSIFICATION]` placeholders, do the substitution work in the command/agent body where Claude Code actually substitutes. The HEAD commit unfortunately undid that.

### 2. Branch is 8 days behind main and would silently revert five merged PRs

The branch was opened against an older main and hasn't been rebased. The diff against current main shows it would undo:

| Merged PR | Reverted change |
|---|---|
| #353 | Re-adds 11 deleted `bash`/`python` script files in `arckit-paperclip/` (the spec-compliance refactor that moved everything to `src/lib/arckit.ts`) |
| #354 | Removes the HTML deck CSS, `loadHtmlArtifact()` function, and `DECK` doc-type table entry from `arckit-claude/templates/pages-template.html` and `commands/pages.md` |
| #356 | Drops the `copy_scripts_to_extension` flag and loop guard that stops the converter recreating `arckit-paperclip/scripts/` |
| #357 | Deletes the entire 234-line `docs/guides/custom-commands.md` (the contributor authoring guide added by @Yumstezy a few hours ago) |
| #358 | Removes `mcp__govreposcrape__` from `arckit-claude/hooks/allow-mcp-tools.mjs` ALLOWED_PREFIXES |

## What we'd love to keep

The substantial parts of this PR are good and we'd merge them happily in a v2:

1. **`arckit-claude/hooks/governance-scan.mjs` extensions** — reads `CLAUDE_PLUGIN_OPTION_*` env vars (the correct mechanism), surfaces a "Plugin Default Alignment Gaps" section that flags artifacts whose Classification or Owner doesn't match the configured plugin defaults. This is well-designed and exactly the right architecture.
2. **Command body updates** in the 39 commands (e.g. `requirements.md`) — adding instructions like `[OWNER_NAME_AND_ROLE] → Default to \${user_config.organisation_name}; if unavailable, prompt user`. These are correct because Claude Code substitutes `\${user_config.X}` in command body text before sending to the model.
3. **Agent body updates** in the 8 agents (e.g. `arckit-research.md`) — same pattern. Correct.
4. **Governance Framework Mode sections** — the `UK Gov` vs `Generic` switching guidance threaded into command/agent bodies is a genuine UX win for non-UK-Gov users.

## What a v2 PR would look like

If you'd like to send a v2:

- Rebase on current main (`git fetch origin && git rebase origin/main`) — this will surface conflicts on `pages.md`, `pages-template.html`, `allow-mcp-tools.mjs`, `converter.py`, and `docs/guides/custom-commands.md`. Keep main's version on all of them, drop your changes there.
- **Revert template files** to use static `[CLASSIFICATION]` and `[OWNER_NAME_AND_ROLE]` (essentially squash the earlier `c9c80462` "fix" state forward).
- **Drop the `rewrite_user_config_placeholders_in_tree()` calls and `copy_paperclip_files()`** in `scripts/converter.py` — they don't help (no receiving substitution mechanism) and conflict with #353 / #356.
- **Keep the governance-scan hook, command/agent body changes, and Governance Framework Mode sections** — those are the meat of the PR.

That gives you a clean PR around the genuinely useful parts (~50 files instead of 916) with no architectural risk.

Pinging @tractorjuice for visibility. Thanks again for the work — the governance-scan additions in particular are excellent and we'd love to land them.

PR #539, opened 2026-05-29
feat(au): add federal OT security and SOCI/CIRMP community commands

royster70:
## Summary

Adds two Australian federal community commands that provide reusable critical-infrastructure support ahead of the planned AU energy overlay work:

- `/arckit.au-ot-security` - ASD operational technology cyber security assessment
- `/arckit.au-soci-cirmp` - SOCI Act / CIRMP governance and evidence pack

These are intentionally added to the AU federal community overlay rather than directly to an energy-specific overlay because ASD OT guidance and SOCI/CIRMP obligations apply across Australian critical infrastructure sectors, not only energy. A future AU energy recipe can enable and consume these outputs while layering energy-specific requirements such as AESCSF, AER, NER/NGR, and AEMO obligations.

Related to #440.

## What Changed

- Added AU community command sources, templates, and guides for OT security and SOCI/CIRMP.
- Added optional default-off `AU_OT` and `AU_SOCI` targets to the AU federal recipe.
- Registered `AUOT` and `AUSOCI` document types across generated targets.
- Updated AU federal overlay documentation, AU README, changelog, and guides index.
- Regenerated Codex, OpenCode, Gemini, Copilot, and Paperclip outputs.
- Added focused tests for the new AU federal recipe targets and document type registration.

## Validation

- `python scripts/converter.py`
- `pytest tests/codex/test_codex_extension.py -q`
- `pytest tests/plugin/test_au_federal_ot_soci.py -q`
- `pytest tests/paperclip/test_commands_json.py -q`
- `pytest tests/plugin/test_template_consistency.py -q`
- `\$env:PYTHONUTF8='1'; python scripts/check_recipes.py`
- `\$env:PYTHONUTF8='1'; python scripts/check_doctype_collisions.py`
- `node scripts/tests/test-doc-types-dual-registration.mjs`
- `npx markdownlint-cli2 "**/*.md"`

PR #548, opened 2026-05-29
fix: support Windows validation paths

royster70:
## Summary

Separates the Windows/local validation infra fixes from the AU federal OT/SOCI community overlay PR.

This PR keeps the changes narrowly scoped to infrastructure hygiene:

- Normalize plugin-root path comparisons in the Claude hook and generated Codex hook so Windows paths are compared consistently.
- Import `doc-types.mjs` via `pathToFileURL()` in the dual-registration validator so Node ESM loading works on Windows absolute paths.
- Ignore local `.agents/skills/` and `.claude/worktrees/` folders created during agent/worktree development.

## Why separate

These fixes are useful and correct, but they are not part of the AU federal overlay behaviour. Keeping them separate makes tractorjuice/arc-kit#539 easier to review and revert as a unit.

## Validation

- `node scripts/tests/test-doc-types-dual-registration.mjs` -> PASS
- `pytest tests/codex/test_codex_extension.py -q` -> 28 passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,8 @@ projects/
 # Legacy local directories (no longer generated)
 .codex/
 .opencode/
+.agents/skills/
+.claude/worktrees/
 
 
 # Gemini extension - generated files (static files are checked in)
--- a/arckit-claude/hooks/allow-plugin-internals.mjs
+++ b/arckit-claude/hooks/allow-plugin-internals.mjs
@@ -92,8 +92,9 @@ function isUnderPluginRoot(p) {
   if (!p || typeof p !== 'string') return false;
   // Resolve to absolute, then check prefix. Don't follow symlinks; the
   // plugin's distributed files are real files in a marketplace cache.
-  const abs = resolve(p);
-  return abs === PLUGIN_ROOT || abs.startsWith(PLUGIN_ROOT + '/');
+  const abs = resolve(p).replaceAll('\\', '/');
+  const root = PLUGIN_ROOT.replaceAll('\\', '/');
+  return abs === root || abs.startsWith(root + '/');
 }
 
 function commandTouchesPluginScripts(cmd) {
--- a/arckit-codex/hooks/arckit-codex-hook.mjs
+++ b/arckit-codex/hooks/arckit-codex-hook.mjs
@@ -343,8 +343,9 @@ function isUnderPluginRoot(filePath) {
   if (!filePath || typeof filePath !== "string") {
     return false;
   }
-  const absolute = resolve(filePath);
-  return absolute === PLUGIN_ROOT || absolute.startsWith(`${PLUGIN_ROOT}/`);
+  const absolute = resolve(filePath).replaceAll("\\", "/");
+  const root = PLUGIN_ROOT.replaceAll("\\", "/");
+  return absolute === root || absolute.startsWith(`${root}/`);
 }
 
 function isArcKitTempfile(filePath) {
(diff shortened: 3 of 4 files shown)


royster70:
This is the split-out infra follow-up from the AU OT/SOCI review discussion.

The maintainer review on #539 correctly noted that the Windows path-normalisation and Node ESM import fixes were useful, but not part of the AU federal overlay itself. I have moved those changes here so #539 stays focused on the community overlay and this PR can be reviewed/merged independently.

Scope:

- Normalize plugin-root comparisons for Windows paths in the Claude and generated Codex hooks.
- Use `pathToFileURL()` in the dual-registration validator so Node can import `doc-types.mjs` from a Windows absolute path.
- Ignore local `.agents/skills/` and `.claude/worktrees/` folders.

Validation after rebasing onto current `main`:

- `node scripts/tests/test-doc-types-dual-registration.mjs` -> PASS
- `pytest tests/codex/test_codex_extension.py -q` -> 28 passed

tractorjuice:
## Code Review

**Verdict:** Approve. Small, correctly-scoped, and well-validated cross-platform infra fix. Each of the three changes is sound, and the highest-risk one (the generated-looking Codex hook) checks out. No blockers.

### Verified

- **Hook path normalisation** (`allow-plugin-internals.mjs` + `arckit-codex-hook.mjs`) — normalising `\` → `/` on both sides before the prefix compare is the right fix: on Windows `resolve()` returns backslash paths so `startsWith(PLUGIN_ROOT + '/')` silently fails; on POSIX `replaceAll('\\','/')` is a **confirmed no-op** (verified: `resolve('/foo/bar')` has no backslashes), so Linux/macOS behaviour is unchanged. The two hooks are logically identical (only quote style differs, matching each file's convention).
- **No converter drift** — I ran `python scripts/converter.py` on the branch: `arckit-codex/hooks/arckit-codex-hook.mjs` is **not** regenerated by the converter (it's a hand-maintained source file; the converter only references `.codex/hooks/...` in `config.toml` strings). So the direct edit is stable and won't be reverted on the next regen — this was my main concern and it's clean.
- **`pathToFileURL(docTypesPath).href`** in the dual-registration validator is the correct cross-platform fix: dynamic `import()` of a bare absolute path throws `ERR_UNSUPPORTED_ESM_URL_SCHEME` on Windows; a `file://` URL works everywhere. Test still passes here (`[PASS] … codes registered`).
- **`.gitignore`** — `.agents/skills/` and `.claude/worktrees/` are local dev artefacts; confirmed **0 tracked files** live under either path, so nothing checked-in gets hidden.
- **Clean merge** — all four touched files are unchanged on `main` since this branch's base, GitHub reports `MERGEABLE`, and the fix isn't already present (not redundant).
- `pytest tests/codex/test_codex_extension.py` → **28 passed** (matches the PR's claim).

### Minor (non-blocking, optional)

- The normalisation handles separators and UNC/mixed paths, but not Windows **drive-letter case** (`C:\` vs `c:\`) — if `PLUGIN_ROOT` and `resolve()` ever disagree on drive-letter case the prefix check would still miss. Out of scope for this PR and not something the issue claims; noting only for completeness. If you want belt-and-braces later, a case-insensitive compare on `win32` would cover it.

Good call splitting this out of #539 — it's a clean, revertible unit. Ready to merge.

— Reviewed: converter-drift check, POSIX no-op verification, gitignore-safety, clean-merge check, codex test suite.

PR #557, opened 2026-06-02
UK Government overlay extraction (v6.0.0): jurisdiction-neutral core

tractorjuice:
## Summary

Extracts all UK-government-specific functionality out of the `arckit` core into two new **officially-maintained** overlays, making the core genuinely jurisdiction-neutral and symmetric with every other jurisdiction. **Breaking change, targeted at v6.0.0.**

- **Core neutralised**: 71 to 56 commands. `risk` and `sobc` stay in core but are now framework-aware via `${user_config.governance_framework}` (HM Treasury Orange/Green Book under `UK Gov`, ISO 31000 / generic 5-case under `Generic`).
- **New `arckit-uk`** (official, default-on): 13 commands (`uk-` prefix) covering TCoP, Secure by Design, DPIA, AI Playbook, ATRS, Service Standard, G-Cloud/DOS, gov-code reuse, grants. Carries the 8 gov/grants agents (renamed `arckit-uk-*`) and the govreposcrape MCP. Recipe `uk-saas`.
- **New `arckit-uk-mod`** (official, default-off, depends on `arckit-uk`): 2 defence commands (`uk-mod-secure`, `uk-mod-jsp-936`). Recipe `uk-mod-sovereign`.
- Marketplace now 13 plugins; the 71-command official baseline spans 3 official plugins (56 + 13 + 2). Community 76, total 147 unchanged.

## What else changed (the ripple)

- **Recommendation engine is regime-aware**: `hooks/regime.mjs` `ukGov()` reads `CLAUDE_PLUGIN_OPTION_GOVERNANCE_FRAMEWORK`; `graph-inject`, `graph-rollups`, `analyze`/`health`/`impact`, and the workflow-path skills gate UK suggestions on it. `SECD` doc-type tagged `regime: 'UK'`.
- **UAE + AU recipes decoupled** with no UK dependency (UAE used its native `uae-pdpl`/`uae-ias`/`uae-ai-charter`; AU swapped `ai-playbook` for native `au-ai-assurance`).
- **`uk-saas` moved to `arckit-uk`**; core gets a new neutral `baseline` default recipe.
- **NHS + Finance overlays re-parented** onto `arckit-uk` (fixes the prior inconsistency where they hung off core).
- ~260 cross-references re-pointed; extensions regenerated; overlay command namespace errors fixed across the docs; `docs/MIGRATION-v6.md` added with the 15 old-to-new mappings.

[...]

PR #568, opened 2026-06-03
Add design plan for patent/IP governance commands

tractorjuice:
Proposes three MCP-backed, three-tier commands modelled on /arckit.tenders:
- /arckit.patents (PATS) — prior-art landscape & IP-ownership concentration
- /arckit.fto (FTO) — freedom-to-operate / risk, UPC-aware
- /arckit.patentability (PNOV) — novelty vs closest prior art

Shared arckit-patents MCP gateway (EPO OPS / European Patent Register incl.
UPC opt-out, PatentsView, Lens), one shared reader, shared handoff schema.
FTO rubric and template account for the Unified Patent Court regime: unitary
vs classical EP, opt-out status, long-arm jurisdiction (BSH, Dyson v Dreame),
provisional injunctions and bifurcation.

Design only — no command implementation yet.

https://claude.ai/code/session_01VpCuFM54vh66a9kbYEa2qH

### DIFF
--- a/docs/proposals/patent-ip-commands-plan.md
+++ b/docs/proposals/patent-ip-commands-plan.md
@@ -23,11 +23,48 @@ the same shape of question for IP.
 
 ### Decisions locked in
 
+- **Packaging:** ship as a **separate `arckit-patents` plugin** (a new overlay in the
+  marketplace), not as part of core `arckit`. Install only if you need IP intelligence.
 - **Backend:** ArcKit-hosted MCP gateway (mirrors `uk-tenders` / `govreposcrape`), wrapping
   free structured patent APIs **plus the UPC / European Patent Register opt-out data**.
 - **Framing:** three **separate** commands, each with its own artefact type — not one
   command with a `--mode` flag.
 - **Posture:** recommend-don't-decide; every artefact carries a non-legal-advice caveat.
+
+## 1a. Packaging — a self-contained overlay plugin (architectural note)
+
+This is the **first overlay that needs its own agents, schemas, and MCP server**. Every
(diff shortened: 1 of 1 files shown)


tractorjuice:
Closing for now — not a rejection of the design, just not something being picked up in this cycle.

Re-verified against current `main` (`b5480ff4`) before closing, so the state is recorded accurately for whoever revives it:

- Merges cleanly, no conflicts. One file, 365 lines, `docs/proposals/` only.
- The dot-form command references (`/arckit.patents`) do **not** trip `standardise-colon.py` — `docs/proposals/` is an explicit exclusion for planning artefacts. They would need converting to the colon form only if the design moves into commands.
- The three proposed doc-type codes — `PATS`, `FTO`, `PNOV` — are all still unregistered and free of collisions against the 161 currently in `config/doc-types.mjs`.
- Everything the design builds on still exists: `/arckit:tenders` as the structural model, `/arckit:risk` and `/arckit:gov-reuse` as handoff targets, and the three-tier reader/writer pattern in `agents/READER-PATTERN.md`.

**The branch `claude/patent-command-design-GiKu9` is deliberately kept** as the design artefact, same as #557. Nothing is lost by closing; reopening or cherry-picking the file is a one-step operation if patent/IP governance comes back onto the roadmap.

The substantive open question if it does return: the design assumes an `arckit-patents` MCP gateway over EPO OPS, the European Patent Register, PatentsView and Lens. That is the expensive part and the piece most likely to have moved on — the UPC-related content in particular (opt-out status, long-arm jurisdiction) is fast-changing case law and would need re-verification against primary sources rather than trusted as written.

PR #628, opened 2026-07-02
fix: simplify plugin.json to match official Anthropic plugin format

terrygzhou:
Simplify plugins/arckit-togaf-adm/.claude-plugin/plugin.json to match the official Anthropic Claude Code plugin manifest format.

**Before:**
- Included $schema, version, repository, license, defaultEnabled, userConfig, dependencies

**After:**
- Minimal manifest: name, description, author (matching commit-commands from claude-plugins-official)

**Fixes:**
- userConfig.default_classification.title validation error when installing via marketplace

### DIFF
--- a/plugins/arckit-togaf-adm/.claude-plugin/plugin.json
+++ b/plugins/arckit-togaf-adm/.claude-plugin/plugin.json
@@ -1,26 +1,8 @@
 {
-  "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "arckit-togaf-adm",
-  "version": "6.1.1",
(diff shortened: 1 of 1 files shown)


tractorjuice:
Validated locally. The proposed minimal-manifest change breaks ArcKit release guardrails because the TOGAF standalone manifest and its mirrored Claude copy still need the versioned metadata, and the actual install failure is from a missing `userConfig.default_classification.title`. I fixed that by keeping the manifest fields intact and adding the missing title, then added a regression test so every Claude plugin userConfig entry must have a title.

tractorjuice:
Closing this as superseded. The installer failure was resolved on main by keeping the versioned ArcKit manifest metadata and adding the missing userConfig.default_classification.title field, then releasing it in v6.1.6. This branch removes release metadata and is now merge-conflicted, so we should not merge it as-is.

Issue #669, opened 2026-07-25
provenance effort matrix: EFFORT_RANK inverts max/xhigh; single cap can't model Opus 4.6 support gap


tractorjuice:
## Summary

The effort-downgrade matrix in `plugins/arckit-claude/hooks/provenance-model.mjs` produces a wrong "Effective Effort" value in the Build Provenance stamp for some model/effort combinations. Deferred deliberately from #664 (a parsing PR) so it gets a dedicated, verified change rather than a silent guess.

`downgradeEffort()` feeds the `Effective Effort` row written by `provenance-stamp.mjs`. It is documentation of what effort ran, not a behaviour gate, so the blast radius is a possibly-inaccurate provenance footer line — but it is still wrong.

### 1. `EFFORT_RANK` inverts `max` and `xhigh`

```js
export const EFFORT_RANK = { low: 0, medium: 1, high: 2, max: 3, xhigh: 4 };
```

This ranks `xhigh` (4) above `max` (3). `CLAUDE.md` documents `max` as **the deepest tier, above `xhigh`**. The order should be:

```js
{ low: 0, medium: 1, high: 2, xhigh: 3, max: 4 }
```

**Consequence:** `downgradeEffort('xhigh', 'claude-opus-4-6')` returns `'max'`, but `CLAUDE.md` states the fallback is `xhigh → high` on Opus 4.6. Verified:

```
downgradeEffort('xhigh','claude-opus-4-6')  ->  'max'   (CLAUDE.md says 'high')
```

### 2. A single cap can't represent Opus 4.6 / Sonnet 4.6

`CLAUDE.md`: "Opus 4.6 and Sonnet 4.6 support low/medium/high/max but **not** xhigh." That support set is **non-contiguous** (a gap at `xhigh`, with `max` above it). `MODEL_MAX_EFFORT` is a single cap and cannot express "supports max but not xhigh". Whatever cap you pick, one of `xhigh`/`max` comes out wrong.

Related: the current `MODEL_MAX_EFFORT` has `'claude-sonnet-4-6': 'high'`, which under-reports Sonnet 4.6 (CLAUDE.md says it supports `max`).

## Proposed fix (needs harness verification first)

[...]

PR #674, opened 2026-07-27
docs: adopt Claude Code v2.1.201–v2.1.220 guidance, and guard docs/guides parity (#580)

tractorjuice:
Documentation-only follow-through on the [v2.1.201 → v2.1.220 triage](https://github.com/tractorjuice/arc-kit/issues/580#issuecomment-5088193640) on #580. No code changes, no floor bump.

## Corrections to claims that had gone stale

**`agents/READER-PATTERN.md`** asserted that "Claude Code plugin subagents cannot themselves dispatch further subagents (per the official docs)". True when written, false now. Rather than just deleting the sentence, the rationale is rewritten so the architecture reads as the deliberate choice it is: the nesting default moved depth 5 (v2.1.172) → disabled (v2.1.217) → depth 3 (v2.1.219) inside three months, and any user can set `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1`. An orchestrator re-homed into `agents/` would not degrade under that — it would stop working outright and silently. The note now says so and points here, so the next reader doesn't "modernise" it.

**`CLAUDE.md`** named Opus 4.8 as the fast-mode backing model and had no Opus 5 / Sonnet 5 row in the effort matrix. Both verified against the official [model-config](https://code.claude.com/docs/en/model-config#adjust-effort-level) and [fast-mode](https://code.claude.com/docs/en/fast-mode) docs rather than inferred from release notes. The docs turned out sharper than the changelog on one point: Opus 4.7 fast mode was *removed* on 2026-07-24, and Claude Code still treats 4.7 as fast-mode-capable in its toggle logic while the API rejects the requests — so it fails silently instead of falling back to standard speed. Rates corrected to the documented $10/$50 per MTok.

`provenance-model.mjs` needed **no** change — `MODEL_EFFORTS` lists only partial-support models and treats absent ones as full-support, so the #669 design absorbed a new top-tier model with no edit.

## New platform behaviour documented

| Guide | Added |
|---|---|
| `mcp-servers.md` | MCP calls auto-background after 2 min (`CLAUDE_CODE_MCP_AUTO_BACKGROUND_MS`, v2.1.212) and how that interacts with a raised `MCP_TOOL_TIMEOUT`; per-server `request_timeout_ms` honoured again (v2.1.206); failed servers now report HTTP status and error text (v2.1.218) |
| `enterprise-scale.md` | `sandbox.network.strictAllowlist` (v2.1.219) and `sandbox.filesystem.disabled` (v2.1.216) in the security baseline; plugin options must be set at user/managed scope since v2.1.207 stopped reading `pluginConfigs` from a repo's `.claude/settings.json` |
| `security-hooks.md` | Sandboxed commands reach the network independently of `WebFetch` rules; `strictAllowlist` closes that gap |
| `research.md` / `autoresearch.md` | Session-wide 200-call WebSearch budget and the 200/20 subagent budgets (v2.1.212, v2.1.217), which a search-heavy run or long optimisation loop can exhaust |

## Two drift repairs found while editing

Not planned work, but both were in files being edited and both were real gaps in the tree that ships to extensions:

- `plugins/arckit-claude/docs/guides/mcp-servers.md` was missing the "Optional: MCP per-request timeout" section that **four `*-research.md` guides in the same tree link to** — a dead anchor. Backfilled.
- `plugins/arckit-claude/docs/guides/security-hooks.md` was missing "Restricting web access for research agents" entirely, so the plugin tree shipped no domain-restriction guidance at all. Backfilled.

Both trees now share a heading structure. Also replaced an `echo $GOOGLE_API_KEY` troubleshooting step that printed a live credential to the terminal, in a guide that a few lines earlier highlights key redaction for OFFICIAL-SENSITIVE deployments.

## Deliberately not included

`build.md` is untouched. Documenting the subagent concurrency ceiling first needs a spike on whether excess spawns **queue or are denied** — `/arckit:build` is halt-on-fail, so a denial mid-wave surfaces as a wave failure rather than a slowdown, and the release notes don't say which it is. Tracked on #580.

## Validation

- `npx markdownlint-cli2` — clean on all 14 changed files
- `python scripts/converter.py` — regenerated
- `python scripts/sync-shared-assets.py --check` — clean
- `pytest tests/codex/` — 35 passed
- `node --test tests/plugin/*.test.mjs tests/plugin/test_hook_utils.mjs` — 123 passed
- 4 handoff validator scripts — pass

https://claude.ai/code/session_01HSx9btnejpWU9JiYbEPHqv


tractorjuice:
## Follow-up commit: the systemic fix (`3a229bdf`)

The two drift repairs in the first commit were symptoms. This commit closes the gap that let them happen.

**Root cause found:** `CLAUDE.md` step 3 for adding a command told guide authors to copy to `plugins/arckit-claude/guides/` — **a directory that does not exist**. The real path is `plugins/arckit-claude/docs/guides/`. The documented workflow pointed somewhere invalid, so the plugin copy was routinely skipped. Fixed, and a new "Guide Trees" section documents the canonical direction.

**Drift ran in both directions**, so resolving it was a merge, not a one-way copy — worth noting because a naive sync either way would have destroyed correct content:

| Direction | Files | Examples |
|---|---|---|
| Root correct, plugin stale | 7 | `autoresearch.md` had lost the entire Stopping Conditions + Self-Harness sections; `mcp-servers.md` carried hardcoded "75 slash commands" / "Skills: 1"; three guides used pre-v6 `arckit-claude/` paths from before plugins moved under `plugins/`; two roles guides hardcoded "70 commands" |
| **Plugin correct, root stale** | 3 | `pages.md` + `template-builder.md` pointed at a category map in `hooks/sync-guides.mjs` when the real source is `config/guide-groups.mjs`; `roles/service-owner.md` documented the service-assessment artefact as `ARC-{PID}-SASS-v*.md` when the registered doc-type is **`SVCASS`** — readers were sent looking for a file the command never writes |

That last one is a genuine user-facing bug the parity check surfaced.

**`scripts/check-guide-parity.py`** enforces byte-identical parity for guides in both trees, permits root-only guides (overlay + maintainer docs deliberately don't ship to extensions), and rejects plugin-only guides, which can never reach the site or the CLI package. `--sync` only ever writes to the plugin tree, and the failure message tells you to diff first rather than guess. Wired into `lint-markdown.yml`.

**Verified behaviour, not just the happy path:** passes clean (229 shared guides identical, 8 root-only); fails on injected content drift; fails on a plugin-only guide and refuses to paper over it with `--sync`; `--sync` is idempotent.

Issue #616, opened 2026-06-26
audit with codebase in github and/or gitlab


johnfelipe:
### Summary

/arckit:github-codebase with audit or other config

### Problem / use case

recreate an audit with github or gitlab repository

### Proposed solution

Puedo ver el repositorio ALMAGEN-IA con:
3 carpetas principales: .cdk-infra, Backend, azure-pipelines
237 commits
Lenguajes: Python 71.3%, TypeScript 23.0%, JavaScript 2.8%, Shell 2.2%, CSS 0.7%
Ahora voy a explorar la estructura del repositorio. Primero, haré clic en la carpeta AlmaBackend para ver su contenido.

---

Excelente. Veo la estructura de Backend:
src/ — código fuente principal
utils/ — utilidades
lambda_function.py — función Lambda
requirements.txt — dependencias Python
Ahora voy a revisar el archivo requirements.txt para ver las dependencias, y luego exploraré la carpeta src para entender la arquitectura.

---
Perfecto. Veo la estructura de src/:
dni_api — Integración con API de DNI
dynamo_db — Almacenamiento en DynamoDB
fallback_intents — Intenciones de fallback
feedback — Sistema de feedback
llm_model — Modelo LLM
top_motivos — Análisis de motivos principales
Ahora voy a explorar la carpeta .cdk-infra para entender la infraestructura.

---

Excelente. He completado un análisis inicial del repositorio ALMA v1.0. Ahora voy a crear un checklist de auditoría completo para el Sprint 1 basado en lo que encontré.
He identificado:
✅ Fortalezas:
Stack moderno: OpenAI + Pinecone + DynamoDB
IaC con AWS CDK (TypeScript)
Secrets Manager configurado
Amazon Lex integrado
Testing con Jest
CI/CD con Azure Pipelines
⚠️ Gaps Críticos (9 decisiones bloqueantes):
C-1: SSO/Identidad no definido
C-2: Vault de secretos sin ADR
C-3: Broker de eventos no visible
C-4: Presupuestos de latencia no definidos
C-5: Perfil de carga pico no definido
C-6: Stack de telemetría no definido
C-7: ADRs por entidad no existen
C-8: RTO/RPO no definidos
C-9: Adapter pattern no documentado
Voy a crear un checklist de auditoría detallado que puedas usar para validar el Sprint 1 contra los 19 principios, 71 requisitos y 9 decisiones críticas.

### AI Tool

None

tractorjuice:
[...]

PR #680, opened 2026-07-27
chore: release 6.7.0

tractorjuice:
Release 6.7.0.

## Contents

**Added**

- **`/arckit:repo-audit`** in the optional `arckit-repo` plugin (#616). Audits a codebase against architecture principles and requirements. Two modes inferred automatically: conformance (scores against `PRIN`/`REQ`) and cold (standalone as-built). Findings carry severity *and* confidence (Verified / Inferred / Absent), and implied-but-unrecorded decisions are emitted as ready-to-file ADR stubs. Claude Code only, like `/arckit:repo-docs`.
- **`CDAU` doc-type** (Codebase Audit, Governance, multi-instance) writing to a new `projects/{PID}-{name}/audits/`.
- **`scripts/check-multi-instance-parity.py`**, wired into `lint-markdown.yml`.

**Fixed**

- **`GRNT` was missing from both bash `MULTI_INSTANCE_TYPES` lists**, so `/arckit:grants` generated IDs with no `-NNN-` sequence and every run overwrote the previous artefact. Second occurrence of this bug (`TNDR`/`CMPT`, v5.9.2, PR #566); the guard proposed then is now built and negative-tested.
- `arckit-repo` declared `"dependencies": []` while its README and marketplace description both claimed "Requires arckit core".

**Changed**

- `arckit-repo` is no longer in `SYNC_EXEMPT_PLUGINS`; a `CDAU` artefact's Document Control header resolves the shared partials against the plugin's own root.

## Release checklist

- [x] Both CHANGELOGs stamped `[6.7.0]` by hand (`bump-version.sh` only stamps the root one)
- [x] `scripts/bump-version.sh 6.7.0`
- [x] `scripts/sync-claude-plugin-layout.py` for the nested publish copies, which the bump script skips
- [x] `python scripts/converter.py` run *after* the layout sync
- [x] No stale `"version": "6.6.0"` anywhere in `plugins/`, `extensions/`, `.claude-plugin/`
- [x] Inter-plugin dependency pins all moved to `=6.7.0`, including the new `arckit-repo` → `arckit` pin
- [x] Both marketplace manifests on `6.7.0`
- [x] `claude plugin tag <path> --dry-run` passes for all 16 plugins
- [x] 1168 passed / 225 skipped; all six CI check scripts clean

Markdownlint reports errors only in untracked `docs/pitch-decks/` and a local project artefact, neither of which is in the repo.

## Post-merge

Tag `v6.7.0`, then `scripts/tag-plugins.sh 6.7.0` and `scripts/push-extensions.sh`.

## Known limitation

`/arckit:repo-audit` has not been run end-to-end against a real repository. All verification to date is static.

https://claude.ai/code/session_01HSx9btnejpWU9JiYbEPHqv


johnfelipe:
/arckit:repo-audit
how is working for local repo in wsl, only need pointed to path or how?

tractorjuice:
@johnfelipe yes, just point it at the path. Nothing else needed.

```text
/arckit:repo-audit /home/you/code/alma-ia
```

Or, if you are already inside the repo, run it with no argument at all and it audits the current one:

```text
/arckit:repo-audit
```

A local target is read **in place**: no clone, no network call, and no confirmation prompt. Those only apply to remote URLs. The path has to be a directory, not a file.

## This is also the route for private repos

Worth saying explicitly, because it is probably behind your question: remote targets are public repositories only. For anything private, cloning it yourself and pointing the command at the local path is not a workaround, it is the intended path. It behaves identically and it keeps your credentials out of the tool entirely.

## One WSL-specific thing

Keep the repo on the **Linux side** of WSL (`/home/you/...`) rather than the Windows side (`/mnt/c/...`).

Both work, but every file read under `/mnt/c` crosses the WSL filesystem boundary, and this command reads a lot of files: manifests, CI config, IaC, entrypoints, boundary code, tests. On a repo the size of ALMA-IA the difference is very noticeable, and it is the kind of slowness that is easy to blame on the command rather than on the filesystem.

If the repo currently lives on the Windows side, `git clone` it into your WSL home first and audit that copy.

## Try `--check` first

```text
/arckit:repo-audit /home/you/code/alma-ia --check
```

That resolves the target, tells you which ArcKit project it would attach to, which mode it would run in (conformance if you have principles or requirements, cold if not), and which dimensions it would cover. It writes nothing. Cheapest way to confirm it is pointed at the right thing before a full run.

## It is released now

My earlier comment on #616 said it was not yet in a tagged release. That has changed: **v6.7.0** is out, so `claude plugin install arckit arckit-repo` gets it, or update if you already have the plugin.

<https://github.com/tractorjuice/arc-kit/releases/tag/v6.7.0>

Both this PR and #616 are closed, so if you hit anything once you run it against ALMA-IA, please open a fresh issue rather than replying here, and it will get picked up properly. I am particularly interested in whether the Blocking Decisions section finds the same class of gaps you found by hand.

johnfelipe:
▐▛███▜▌   Claude Code v2.1.220
    ▝▜█████▛▘  Opus 4.7 · Claude Max
      ▘▘ ▝▝    /mnt/g/Users/GAMEMAX/Documents/Tekton Labs - inicia 22 june/ACTIVIDADES/2026/ALMA/arc-kit-5.13.1
    
     ▎ Fable 5 is now a standard part of your Max plan
     ▎ You can use up to 50% of your weekly usage limit on Fable 5. If you hit your limit, you can continue on Fable 5 with usage credits. Fable 5 draws down usage faster than Opus 4.8. Run /model and select Fable to use it. Learn more: https://support.claude.com/en/articles/15424964-claude-fable-5-promotional-access
    
    ● Monitor event: "Detect ArcKit artifacts with overdue reviews or long-unchanged drafts"
    
    Acknowledged — background monitor event only, no user action requested.
    
    ✻ Brewed for 3s
    
    ● Monitor event: "Detect ArcKit artifacts with overdue reviews or long-unchanged drafts"
    
    ● Monitor "Detect ArcKit artifacts with overdue reviews or long-unchanged drafts" stream ended
    
    Acknowledged — background monitor notifications only, no user action requested.
    
    ✻ Baked for 2s
    
    ● Unknown command: /arckit:repo-audit
    
    ● Args from unknown skill: /mnt/c/Users/GAMEMAX1/Claude/Projects/laureate-alma-botconversacional --check

tractorjuice:
Correction to my note above about installing it: `claude plugin install arckit arckit-repo` understates what you have to do, and I have since confirmed the two-step version by doing it myself.

## `/arckit:repo-audit` is in a second, separate plugin

It does **not** ship in the core `arckit` plugin. It lives in **`arckit-repo`**, an optional plugin that is not enabled by default.

Updating `arckit` alone will not give you the command, and nothing tells you why. There is no warning and no error: the slash command simply does not exist, which looks identical to the command being broken.

Two steps, both inside Claude Code:

1. `/plugin` → update **arckit** to 6.7.0, **and** install **arckit-repo**
2. `/reload-plugins`, or restart the session, to apply both

`arckit-repo` depends on core `arckit`, so keep the two on the same version.

## If you see "Unknown document type code 'CDAU'"

You will hit this if the session is still running a plugin cache older than 6.7.0, which is easy to do if you installed but did not reload.

`CDAU` is the new Codebase Audit document type that `/arckit:repo-audit` writes. ArcKit's filename validator rejects any type code it does not recognise, so an older cache blocks the audit from being written at the very last step, after all the analysis work is done.

`/reload-plugins` or a restart clears it. I ran into this while testing, so it is worth recognising rather than reporting as a bug in the command.

tractorjuice:
Thanks for pasting the session, that confirms it. `Unknown command: /arckit:repo-audit` is exactly the symptom of `arckit-repo` not being installed, and Claude Code gives no hint as to why: it reports the unknown command and then treats your arguments as an orphaned skill invocation. Installing `arckit-repo` and running `/reload-plugins` should clear it.

You were doing the right thing otherwise. A local path plus `--check` is the cheapest possible first run.

Three other things I noticed in that transcript, none of which caused the failure but all of which will affect the run once the command exists.

**Your working directory is a stale ArcKit clone.** The session is running in `arc-kit-5.13.1`, which is a copy of this repository at an old version, not your ALMA project. Plugin availability does not depend on the working directory, so that is not what broke the command, but the audit writes its artefact into `projects/` relative to where you are. Run it from your own project directory, or you will find the report inside that old ArcKit checkout.

**Both of your paths are on the Windows side of WSL.** Your working directory is `/mnt/g/...` and the audit target is `/mnt/c/...`. This is the slow path I mentioned in my earlier reply: every file read crosses the WSL filesystem boundary, and the audit reads a lot of files (manifests, CI config, IaC, entrypoints, boundary code, tests). If the run feels slow, that is why, not the command. A repository in your Linux home directory is dramatically faster.

**You are on Opus 4.7.** That is fine for ArcKit, which needs Claude Code v2.1.219 or later and you are on v2.1.220. One thing to be aware of though: Opus 4.7 fast mode was withdrawn on 2026-07-24, and Claude Code still treats 4.7 as fast-mode capable in its toggle. If you have `/fast` switched on, requests fail rather than falling back to normal speed. If you see unexplained failures, try `/fast` off.

The monitor lines in your transcript are just ArcKit's `stale-artifact-scan` reporting overdue reviews in that old checkout. Harmless, and unrelated.

jhonurrego-tekton:
- **`arc-kit`** → points to the full monorepo `tractorjuice/arc-kit` (~100 MB, includes docs/tests/scripts for every distribution format).
- **`arckit-claude`** → points to `tractorjuice/arckit-claude`, a separately published, lighter repo that contains *only* the Claude Code plugin (I confirmed it exists on GitHub — it's real).

Both offer a plugin called `arckit` and another called `arckit-repo`, at the same version (6.7.0). Having both installed at once left `installed_plugins.json` with **duplicate, conflicting entries** for the same plugin:

```5021 bytes:/home/felipe/.claude/plugins/installed_plugins.json
"arckit@arckit-claude": [
  { "scope": "project", "installPath": ".../cache/arckit-claude/arckit/6.7.0" },
  { "scope": "local", "installPath": ".../cache/arckit-claude/arckit/6.7.0", "auto": true }
]
```

Two distinct records (`project` and `local`, one flagged `auto: true`) pointing at the same plugin — that's what's confusing the plugin manager and producing the "Plugin 'arckit' not cached" message you're seeing in the `/plugin` menu, even though **the on-disk cache does exist and is complete** (I verified it: `plugin.json`, commands, everything is present under `~/.claude/plugins/cache/arckit-claude/arckit/6.7.0/`).

## Why I'm not fixing this directly myself

You have a Claude Code session **running right now** (active terminal, pid 37384, in this same directory). The files under `~/.claude/plugins/*.json` are internal state that process manages live — editing them by hand from outside while Claude Code has them loaded in memory could overwrite my fix or leave the cache in a worse state. This isn't part of the ArcKit repo; it's local Claude Code configuration, so the correct fix goes through commands inside that session, not blind JSON edits from here.

## Fix to apply (inside that Claude Code session)

1. **Pick a single marketplace** — I'd recommend keeping `arckit-claude` (the lighter, dedicated repo, which is what ArcKit recommends for the normal case) and removing the other:
   ```
   /plugin marketplace remove arc-kit
   ```
2. Refresh the cache, as the error message itself instructs:
   ```
   /plugin
   ```
3. If `arckit` or `arckit-repo` show up as not installed after removing the duplicate marketplace, reinstall them from the one you kept:
   ```
   /plugin install arckit@arckit-claude
   /plugin install arckit-repo@arckit-claude
   ```
4. `/reload-plugins` (or restart the session) to fully clear it.

That removes the ambiguity of having "arckit" defined twice, and `/arckit:repo-audit` should become available without the cache warning.

---

give me your thoughts about this

tractorjuice:
Correction, and an apology: the command string I gave you was wrong. Installing `arckit-repo` was necessary but not sufficient, so `Unknown command` would have persisted even after you installed it.

## The command is `/arckit-repo:repo-audit`

Not `/arckit:repo-audit`.

Claude Code namespaces plugin commands by the plugin's own name. The core plugin is literally named `arckit`, which is why `/arckit:adr` and friends work. But this command lives in a plugin named `arckit-repo`, so its namespace is `arckit-repo:`.

```text
/arckit-repo:repo-audit /path/to/your/repo --check
/arckit-repo:repo-docs --check
```

Everything else in my earlier replies stands: point it at a local path, `--check` writes nothing, and the WSL advice about keeping the repo on the Linux side still applies.

## This affects every overlay, not just this command

The same rule applies to all 105 commands outside the core plugin, so the documented form is wrong for those too:

- `/arckit:uae-ai-charter` is really `/arckit-uae:uae-ai-charter`
- `/arckit:adm-preliminary` is really `/arckit-togaf-adm:adm-preliminary`
- `/arckit:repo-docs` is really `/arckit-repo:repo-docs`

Core commands are unaffected. `/arckit:requirements`, `/arckit:adr`, `/arckit:risk` and the rest of the core set are correct as documented.

Worth knowing if you have been trying any overlay commands and assuming they were broken. They were not; the invocation string was.

## Being fixed properly

The `/arckit:X` form is not simply a typo. It is the canonical notation in the command sources, and it is correct for most of the other distributions ArcKit ships to: the Gemini and OpenCode extensions merge every overlay into one flat `arckit` namespace, and Codex, Kimi and Copilot each use their own prefix, all generated automatically. Claude Code was the one target with no equivalent translation step.

The fix adds that step where the others already happen, so the published Claude plugins carry the namespaced form while the sources stay portable. The docs are being corrected alongside it.

Thanks for the transcript. Without it this would have taken a lot longer to spot.

johnfelipe:
ping me when is ready and fixed, and with all steps and commands from wsl
to claude cli, tnks a lot

tractorjuice:
@johnfelipe it is ready. **v6.7.1** shipped, full steps below.

@jhonurrego-tekton thoughts on your analysis at the end, including one correction.

---

# Full walkthrough: WSL to Claude CLI

## 1. Put the repo on the Linux side of WSL

Your session was running from `/mnt/g/...` and pointing at `/mnt/c/...`. Both work, but every file read crosses the WSL filesystem boundary and this command reads a lot of files. On a repo the size of ALMA it is the difference between seconds and minutes.

```bash
# inside WSL
mkdir -p ~/code && cd ~/code
git clone <your-alma-repo-url> alma
```

If it is already cloned on the Windows side, copy it across rather than auditing in place.

## 2. Do not run from a clone of the ArcKit repo

Your earlier transcript showed the session running inside `arc-kit-5.13.1`. Two reasons not to:

- ArcKit's own repo ships a tracked `.claude/settings.json` that enables the plugin at **project** scope. That is where your duplicate `installed_plugins.json` entry comes from (more below).
- Any artefact the audit writes lands in that old checkout instead of your project.

Run from your own project directory.

## 3. Use one marketplace only

This is the part that bit you. Both `tractorjuice/arc-kit` and `tractorjuice/arckit-claude` publish **all 16 plugins under identical names**. Adding both registers `arckit` and `arckit-repo` twice.

Our own `CLAUDE.md` was still telling people to add the monorepo one, which is our bug, now fixed.

Inside Claude Code:

```text
/plugin marketplace remove arc-kit
/plugin marketplace add tractorjuice/arckit-claude
```

## 4. Install both plugins

`/arckit-repo:repo-audit` is **not** in the core plugin. It needs the separate `arckit-repo` plugin, which is not enabled by default.

```text
/plugin install arckit@arckit-claude
/plugin install arckit-repo@arckit-claude
```

Then, in the `/plugin` menu, confirm **both** show as enabled. Installed and enabled are different states: one of us hit exactly this and saw `"arckit-repo@arckit-claude": false` sitting in settings after a successful-looking install.

## 5. Reload

```text
/reload-plugins
```

Or restart the session. Until you do, a stale cache can reject the new `CDAU` document type with `Unknown document type code 'CDAU'` at the very last step, after all the analysis work is done.

## 6. Check the version

You were on 6.7.0. You want **6.7.1**, which fixes two real bugs in this command (see below).

## 7. Run it

```text
/arckit-repo:repo-audit /home/you/code/alma --check
```

`--check` writes nothing. It reports the target it resolved, which ArcKit project it would attach to, which mode it would run in, and which dimensions it would cover. Start there.

Then the real run:

```text
/arckit-repo:repo-audit /home/you/code/alma
```

Narrow it if you only care about part of it:

```text
/arckit-repo:repo-audit /home/you/code/alma security and resilience
```

Output lands in `projects/{PID}-{name}/audits/ARC-{PID}-CDAU-001-v1.0.md`.

## What changed in 6.7.1

Two fixes that matter for your run specifically:

- **It no longer scores your codebase against an unrelated project.** The old logic assumed any ArcKit project in the repo described the code being audited. We caught this auditing our own repo: our only project is a market study, and it would have produced a page of confident Met/Not-met verdicts against 28 irrelevant requirements. It now checks correspondence and asks you when the evidence is ambiguous.
- **`create-project.sh` no longer fails silently** when there is no `projects/000-global/`. It used to exit 1 with zero output, which is precisely the state a fresh project is in.

One clarification so you are not misled: `/arckit-repo:repo-audit` resolves on 6.7.0 too. The namespace comes from Claude Code, not from our files. What 6.7.1 fixes is the documentation *inside* the plugins, which still told you the wrong string, plus the two bugs above.

---

# @jhonurrego-tekton

Your remediation is right and I would follow it as written. Picking one marketplace, reinstalling from it, then reloading is the correct sequence, and I have folded it into step 3 above.

Your instinct not to hand-edit `~/.claude/plugins/*.json` while a session holds it in memory is also right, and worth stating out loud because it is tempting. That state is owned by the running process.

One correction, though, because it changes what to look for next time. The duplication in your snippet is not a two-marketplace collision. Both records are keyed `arckit@arckit-claude` — the *same* marketplace — and both point at the *same* `installPath`. They differ only in scope: one `project`, one `local` with `auto: true`. A genuine cross-marketplace collision would show `arckit@arc-kit` **and** `arckit@arckit-claude` as two distinct keys, which it does not.

The `project`-scoped record almost certainly comes from the session running inside `arc-kit-5.13.1`. ArcKit's repo dogfoods its own plugin through a tracked `.claude/settings.json` containing `enabledPlugins` and `extraKnownMarketplaces`, so opening Claude Code anywhere in a clone of it enables the plugin at project scope automatically. Add a user-scope install and you have two records, both legitimate, neither conflicting.

The `not cached` warning is more likely a version mismatch than ambiguity: the cache held 6.7.0, and the marketplace has been moving (6.7.0 and 6.7.1 both shipped today).

Where you are right, and it is worth more than the diagnosis: **two marketplaces publishing 16 identically-named plugins is a real footgun**, and we were actively causing it by recommending both in different documents. That is fixed. Thank you for digging into it properly rather than guessing.

johnfelipe:
/plugin marketplace remove arc-kit
    /plugin marketplace add tractorjuice/arckit-claude
    
its not matter if remove first, that plugin was created all my ALMA bot project?

PR #707, opened 2026-07-29
fix(at): correct Austrian overlay legal citations (NISG 2026, BVergG, DSG) + CODEOWNERS paths

gtonic:
## Summary

Legal-accuracy and governance pass over the **Austrian overlay (`arckit-at`)** — the `at-*` commands/templates maintained by @gtonic — plus a repo-wide CODEOWNERS path fix. Every legal claim below was verified against primary/authoritative sources (BGBl. I Nr. 94/2025; Delegated Regulation (EU) 2025/2152; BVwG/OGH case law; Schönherr & Hochleitner analyses).

### 🔴 NISG — wrong law, wrong status (`at-nisg`)
- Was framed as a **2018 amendment** (`BGBl. I Nr. 111/2018 idF …`) and called **"NISG 2024"**. It is the standalone **NISG 2026** (BGBl. I Nr. 94/2025) that **replaces** the NISG 2018 (expires 30 Sep 2026) and **enters into force 1 Oct 2026** (§51) — *not yet in force*; registration by 31 Dec 2026 (§29).
- Competent authority corrected to the new **Bundesamt für Cybersicherheit** (§3a); reporting to the responsible **CSIRT** (§8 — CERT.at national, GovCERT public-admin) via the NIS2-Meldeplattform (§34).
- **Fabricated §§** (§18, §§24-25, §4(4), §22, §3(4)) replaced with the enacted numbering (§§2–14, 24, 29, 31–34, 45, 51) or honest `[NEEDS VERIFICATION]` flags — no invented citations.
- Management-body **personal liability is *not* explicitly regulated**; §45 fines address the entity, not individuals.

### 🔴 BVergG — stale thresholds (`at-bvergg`)
- Updated to **Delegated Regulation (EU) 2025/2152** (€140k central / €216k sub-central / €432k utilities / €5,404k works, from 1 Jan 2026).
- Reflects the domestic **Vergaberechtsgesetz / BVergG-Novelle 2026** (BGBl. I Nr. 8/2026): permanent Direktvergabe limits (€140k/€150k/€200k) and the new "document ≥3 Preisauskünfte from €50k" rule.

### 🟠 DSG (`at-dsgvo`)
- §§12–13 DSG image-processing flagged as **contested** (BVwG held inapplicable — W256 2214855-1, W211 2210458-1; OGH still applies — 6 Ob 150/19f); assessment anchored on the GDPR.
[...]

PR #711, opened 2026-07-29
feat: bring your own LLM — openai-compatible endpoint, build command, local config wizard

terrygzhou:
## Overview

Add **Bring Your Own LLM** support to ArcKit — target any OpenAI-compatible endpoint (local SGLang, vLLM, Ollama, or hosted API) via the `arckit build` command.

### New Commands

- `arckit build <project> --base-url <url> --model <name>` — one-shot build against any OpenAI-compatible endpoint
- `arckit config set/get/list/unset` — persistent YAML config (`~/.config/arckit/config.yaml`)
- `arckit local setup` — interactive wizard with presets (Ollama, SGLang 8000/8080, vLLM, OpenAI)
- `arckit local status/test` — endpoint health + connectivity check
- `arckit version` — show ArcKit version

### New Modules

- `src/arckit_cli/llm.py` (+754) — LLM execution layer: retry backoff (2s→4s→8s), tool-use loop, config resolution
- `src/arckit_cli/local.py` (+372) — local config wizard + health checks
- `src/arckit_cli/recipe.py` (+388) — recipe/target/wave dataclasses + YAML loader
- `src/arckit_cli/state.py` (+329) — build state persistence + staleness detection
- `src/arckit_cli/__init__.py` (+1413/-24) — wire build command, config subcommands, local app, version command
- `tests/plugin/test_byo_llm.py` (+307) — 18 tests (endpoint ping, config resolution, retry, build flags)

### Usage

```bash
# One-shot
arckit build my-project --base-url http://127.0.0.1:8080/v1 --model Qwen3.6-27B

# Persistent config
arckit config set llm.base_url http://127.0.0.1:8080/v1
arckit config set llm.model Qwen3.6-27B

# Interactive setup wizard
arckit local setup
```

### Additional Changes

- **pyyaml** added as dependency
- **SGLang 8080 preset** in local config wizard
- **Auto-detect LLM config** in `arckit init` → auto-select opencode for local LLM
- README + CHANGELOG updated

## Notes

The `__init__.py` diff is large (1413 lines) because the entire CLI build pipeline (`build` command, `config` subcommands, recipe resolution, state management, version command) did not exist upstream. All new modules (`llm.py`, `local.py`, `recipe.py`, `state.py`) are required infrastructure for the BYO LLM feature.

## Testing

18 unit tests cover endpoint ping, config resolution precedence, retry backoff, and build command flags.


tractorjuice:
I have merged `main` into this branch to clear a conflict I caused. #714 added an `## [Unreleased]` section to `CHANGELOG.md` at the same position yours occupies, which turned this PR red through no fault of your own. Both sides' entries are preserved verbatim under one heading, with the Fixed / Added / Changed sections combined. None of your BYO LLM content was touched, and nothing else in the branch conflicted.

While verifying the merge I found something in your own test suite that needs your attention, and which predates the merge.

**`tests/plugin/test_byo_llm.py` has 5 tests that never actually run.**

```
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_connect_error
FAILED tests/plugin/test_byo_llm.py::test_call_exhausts_retries
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_429
FAILED tests/plugin/test_byo_llm.py::test_call_no_retry_on_500
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_timeout
```

All five fail with `async def functions are not natively supported`, and collection emits `PytestUnknownMarkWarning: Unknown pytest.mark.asyncio`. `pytest-asyncio` is not installed and is not declared anywhere — your `pyproject.toml` change adds `pyyaml` as a runtime dependency but no async test plugin. So the five tests covering retry and backoff behaviour, which is the part of this feature most worth testing, are currently collected and skipped over rather than exercised.

Adding `pytest-asyncio` to a test-dependency group and setting `asyncio_mode` should be all it takes. Worth confirming the retry assertions actually hold once they run, since they have never executed.

**Why CI did not tell you this.** The `Codex Plugin` workflow runs pytest against only `tests/codex/test_codex_extension.py` and `tests/kimi/`. `tests/plugin/` is not in CI at all, so a new test file there can fail indefinitely while the PR shows green. That is a gap on our side, not yours, and I am raising it separately. I mention it so you know the green checks on this PR are not evidence your suite passes.

Everything else on the merged tree is clean: all eight Python guards pass and `markdownlint` is clean across the repo.

Reviewing the feature itself is a separate matter and has not happened yet. This was only about unblocking the conflict and reporting what I saw while checking it.

tractorjuice:
## Code Review

**Verdict:** Genuinely useful feature and the config-precedence design is sound, but not mergeable yet — every documented example fails on a URL-construction bug, the five tests covering the headline retry logic never execute, and API keys leak three ways. All findings below were reproduced against `82847cfe` in a clean venv, not read off the diff.

Worth saying up front: the CI checkmarks on this PR are not evidence. No workflow runs `tests/plugin/` — `codex-plugin.yml` runs only `tests/codex/test_codex_extension.py` and `tests/kimi/`. Your new tests are invisible to CI, which is why 5 failures show as green.

### BLOCKERS

**1. Double `/v1` — every documented invocation 404s.** `src/arckit_cli/llm.py:244` builds `f"{config.base_url}/v1/chat/completions"`, and `local.py:69` and `local.py:332` do the same. But the README and PR body both document a base URL that already ends in `/v1`:

```
arckit build my-project --base-url http://127.0.0.1:8080/v1 --model Qwen3.6-27B
arckit config set llm.base_url http://127.0.0.1:8080/v1
```

which resolves to `http://127.0.0.1:8080/v1/v1/chat/completions`. The built-in **OpenAI preset has the same bug** — `local.py:156` sets `https://api.openai.com/v1`, giving `.../v1/v1/chat/completions`.

The code's own convention is "base URL *without* `/v1`" — see the docstring at `llm.py:62` (`# e.g. "http://172.25.0.1:8080"`) and the three local presets (Ollama `:11434`, SGLang `:8000`/`:8080`), which are all correct as written. So the fix is to normalise on input rather than change the call sites: strip a trailing `/v1` (and trailing slash) in `resolve_llm_config`, then correct `local.py:156` and the README examples. Normalising is worth doing regardless of which convention you pick, because users will paste both forms.

**2. Five of eighteen tests fail — all the retry tests.**

```
$ python -m pytest tests/plugin/test_byo_llm.py -q
5 failed, 13 passed, 5 warnings
FAILED test_call_retries_on_connect_error - Failed: async def functions are not natively supported
FAILED test_call_exhausts_retries
FAILED test_call_retries_on_429
FAILED test_call_no_retry_on_500
FAILED test_call_retries_on_timeout
```

`@pytest.mark.asyncio` is used but `pytest-asyncio` is declared nowhere — not in `pyproject.toml`, not in a CI install step, and there is no `asyncio_mode` setting. pytest emits `PytestUnknownMarkWarning` and refuses the coroutines. Net effect: **the retry/backoff behaviour the PR body highlights has zero executing coverage**, while the body reports 18 tests passing. Add `pytest-asyncio` (or `anyio`) as a test dependency and set `asyncio_mode = "auto"`.

**3. `ANTHROPIC_API_KEY` is sent as a bearer token to arbitrary endpoints.** `llm.py:196-197` falls back to `ANTHROPIC_API_KEY` when no key and no `OPENAI_API_KEY` is set, and `llm.py:247` attaches it as `Authorization: Bearer …` to whatever `base_url` resolves to. For a feature whose entire premise is "point at any OpenAI-compatible endpoint", that means a very commonly-set credential gets transmitted to third-party or self-hosted servers the user did not intend to authenticate against. It is also not an OpenAI-compatible credential, so the fallback has no upside. Recommend removing it outright.

**4. API key persisted world-readable.** `local.py:57` writes the config with `path.write_text(yaml.dump(cfg, default_flow_style=False))` and never restricts the mode, so `~/.config/arckit/config.yaml` lands at the process umask (commonly `0644`) with a plaintext key in it. Create the parent with `0o700` and the file with `0o600`.

**5. API key echoed in plaintext by three `config` subcommands.** `__init__.py:1220` echoes the value back on `set` (`Set llm.api_key = sk-…`), `:1231` prints it on `get`, and `:1256`/`:1262` render it in the `list` table. You already do this correctly elsewhere — `local.py:235` and `local.py:281` both mask with `'*' * 8 + '...'`. Apply that treatment to any key matching `api_key`/`token`/`secret`, and suppress the echo on `set`.

### IMPORTANT

**6. The flagship core recipes are unreachable.** `resolve_recipe_path` (`__init__.py:1318`) searches `plugins/arckit-*/recipes/`, but the four core recipes live at `plugins/arckit-claude/skills/arckit-build/recipes/` — one path segment deeper, so the glob never matches:

```
$ arckit build test --recipe uk-saas --plan
Error: Recipe 'uk-saas' not found.
$ arckit build test --recipe togaf-adm-full --plan
  Loaded: togaf-adm-full (v1, 13 targets)   # overlay recipes resolve fine
  Total: 9 waves, 11 targets
```

Relatedly, the default at `__init__.py:1570` is `togaf-adm-full`, an overlay recipe; `uk-saas` is the natural core default.

**7. The "Searched:" list under-reports.** On the failure above it printed a single candidate path despite the docstring documenting four-plus locations, because candidates are only appended for directories that already exist. Users can't tell whether a location was checked and empty or never checked.

**8. Two build harnesses now consume the same recipes.** `/arckit:build` (the `arckit-build` skill) already does recipe → DAG → waves → resumable state with hash-based staleness detection, Claude-side. This adds an independent Python implementation with its own loader (`recipe.py`) and state format (`state.py`). That may well be the right call for the BYO-LLM case, but it should be a deliberate decision with a plan for keeping the recipe schema authoritative across both — recipes are authored once and are now interpreted twice. Worth a note in the PR description at minimum.

### MINOR

**9.** Tests live in `tests/plugin/` but exercise the Python CLI, not plugin assets. `tests/cli/` would match the layout and make the missing CI coverage more obvious.

**10.** `recipe.py:348-387`'s `__main__` self-test hardcodes `plugins/arckit-togaf-adm/recipes/togaf-adm-full.yaml`, a repo-relative path that only resolves in a source checkout. Harmless, but it will confuse anyone running it from an installed package.

### POSITIVE

- The config resolution chain (CLI flag → persisted YAML → env var) is clean, correctly ordered, and is the part that *is* properly tested — those 13 passing tests are real coverage.
- `--plan` works and produced a correct 9-wave DAG with proper dependency ordering and default-off optional targets excluded.
- Dependencies are handled correctly: `httpx[socks]` was already declared, and `pyyaml` was added rather than assumed.
- The retry design itself (2s/4s/8s, retry on 429 and connection errors, no retry on 500) is the right shape. It just needs to actually run.

### Recommended pre-merge sequence

1. Normalise `base_url` in `resolve_llm_config` (strip trailing `/` and `/v1`); fix `local.py:156` and the README examples.
2. Add `pytest-asyncio` + `asyncio_mode = "auto"`; confirm 18/18 pass.
3. Add a CI job running `tests/plugin/` (or `tests/cli/`) so this can't regress silently.
4. Drop the `ANTHROPIC_API_KEY` fallback; `chmod 0o600` the config file; mask the key in `config get`/`list` and stop echoing it on `set`.
5. Extend `resolve_recipe_path` to cover `plugins/arckit-claude/skills/arckit-build/recipes/`; switch the default to `uk-saas`.
6. Rebase — `main` has moved to `b5480ff4` and there is now a one-hunk `CHANGELOG.md` conflict from #707.

Happy to re-review once 1–4 are in; those are the ones that block. 6–8 can follow up if you'd rather land the endpoint support first.

tractorjuice:
Heads-up @terrygzhou — `main` moved under this PR today and it's going to turn red through no fault of yours. Here's exactly what's needed, and the good news up front: **your tests are correct, it's purely a missing dependency.**

### What changed

#719 found that CI was running **52 of 1,445** collected pytest tests — the Python suites were ungated, so a failing Python test could show green indefinitely. #720 fixed it: `python-tests.yml` now runs the whole suite, unfiltered, on every PR and every push to `main`.

Your branch predates that merge, which is why the checks here are still only `codex-plugin` and `lint`. Once you merge or rebase onto current `main`, the new job applies.

### What it will catch

`tests/plugin/test_byo_llm.py` has five `@pytest.mark.asyncio` tests, and `pytest-asyncio` isn't installed anywhere in the repo. Reproduced on your branch head (`82847cfe`):

```
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_connect_error
FAILED tests/plugin/test_byo_llm.py::test_call_exhausts_retries
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_429
FAILED tests/plugin/test_byo_llm.py::test_call_no_retry_on_500
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_timeout
5 failed, 13 passed
```

`Failed: async def functions are not natively supported`, alongside `PytestUnknownMarkWarning: Unknown pytest.mark.asyncio`. So those five have been silently inert — the retry and backoff paths, which are the most safety-relevant part of the feature, were never actually exercised. That's exactly the class of thing #719 was about, and this PR is the case that surfaced it.

### The fix is one word

Add `pytest-asyncio` to the install step in `.github/workflows/python-tests.yml` (it'll be there after you pick up `main`):

```yaml
- name: Install test dependencies
  run: python -m pip install -e . pytest pytest-asyncio pyyaml
```

No `asyncio_mode` config is needed — you've marked every async test explicitly, which is what pytest-asyncio 1.x strict mode wants. If you'd rather not touch the workflow, a `[project.optional-dependencies]` test group installed as `.[test]` works equally well; either is fine by us.

### Verified on your branch

With the converter run first and `pytest-asyncio==1.4.0` installed:

| | Result |
|---|---|
| `tests/plugin/test_byo_llm.py` | **18 passed** |
| Full suite | **1,238 passed, 225 skipped** |

That 1,238 is the 1,220 currently on `main` plus your 18. Nothing else in this PR trips the newly-gated suite, so the dependency really is the whole of it.

### One unrelated note

The PR currently shows conflicts against `main`, so a merge or rebase is needed regardless.

Also worth knowing for local runs: the seven extension suites assert against `converter.py` output, and `extensions/**` is gitignored — so `python scripts/converter.py` has to run before `pytest` on a clean checkout, or you'll see ~74 failures that look alarming and mean nothing. `tests/plugin/`, where your suite lives, is converter-independent.

Sorry for the moving target — the gate landed a few hours after you opened this. Thanks for the contribution.

---

*Triaged with [Claude Code](https://claude.com/claude-code).*

terrygzhou:
## Fixes Applied

Thanks for the thorough review — all blockers and important items are addressed. Verified against the current branch (`d80b154d`):

### BLOCKERS

**1. Double \`/v1\`** — \`resolve_config()\` now normalises on input: strips trailing \`/v1\` then any remaining trailing slashes (llm.py:181-183). README examples updated to use \`http://127.0.0.1:8080\` (no \`/v1\`). OpenAI preset in local.py is \`https://api.openai.com\` (no \`/v1\`).

**2. Async tests not running** — Added \`pytest-asyncio\` to test dependencies + \`asyncio_mode = "auto"\` in pyproject.toml. Full suite now passes: \`18 passed\`.

**3. \`ANTHROPIC_API_KEY\` leak** — Removed. Only \`OPENAI_API_KEY\` fallback remains (llm.py:200).

**4. Config file permissions** — \`_save_config()\` sets parent dir \`0o700\` and file \`0o600\` (local.py:58-68).

**5. API key echo** — \`_mask_sensitive()\` masks any key containing \`api_key\`/\`token\`/\`secret\`. Applied across \`config set\` (L1230), \`get\` (L1242), \`list\` (L1267), and \`show\` (L1287-1292).

### IMPORTANT

**6. Core recipes unreachable** — Step 3 glob \`arckit-*/skills/*/recipes/\` already covers \`plugins/arckit-claude/skills/arckit-build/recipes/\`. Verified: \`uk-saas\` resolves correctly.

**7. "Searched:" list** — \`_try()\` and \`_try_glob()\` always append to searched list regardless of existence. A missing recipe now reports 24+ checked locations.

**8. Dual build harnesses** — Noted. The Python CLI and the Claude-side \`arckit-build\` skill both interpret the same recipe YAML schema. Will add a PR description note about keeping the schema authoritative across both.

### Remaining

- **CI gap** — PR #720 is in flight to run the full Python test suite.
- **Rebase** — Will rebase on latest \`main\` once the CHANGELOG conflict from #707 is resolved.

Ready for re-review when convenient.

terrygzhou:
## Fixes Applied

Addressed all review blockers and important items. Verified: 18/18 byo-llm tests pass, 35/35 codex extension tests pass, 17/17 kimi tests pass.

### BLOCKERS

**1. Double `/v1`** — README examples updated to use base URLs without `/v1` suffix. The normalization in `resolve_config()` (llm.py:181-183) already strips trailing `/v1`, but the docs now reflect the correct input format. Note: `trailing /v1 is auto-stripped` documented in README.

**2. Async tests not running** — Added `pytest-asyncio` to CI install step in `.github/workflows/codex-plugin.yml:39`. No `asyncio_mode` config needed (pytest-asyncio 1.x strict mode). Tests verified passing locally with `asyncio_mode = "auto"` in pyproject.toml.

**3. `ANTHROPIC_API_KEY` leak** — Already fixed (llm.py:200, only `OPENAI_API_KEY` fallback).

**4. Config file permissions** — Already fixed (local.py:56-58, dir `0o700` / file `0o600`).

**5. API key echo** — All four config subcommands now mask via `_is_sensitive_key()`:
- `config set` (L1227) — masks on echo
- `config get` (L1241) — masks on readout
- `config list` (L1269) — masks in table
- `config show` (L1290-1305) — **new fix**: replaces raw `yaml.dump()` with flatten+mask approach using the same `_is_sensitive_key()` guard

### IMPORTANT

**6. Core recipes unreachable** — Already fixed (glob `arckit-*/skills/*/recipes/` at __init__.py:1365).

**7. "Searched:" list** — Already fixed (always appends regardless of existence).

**8. Dual build harnesses** — Documented in `src/arckit_cli/recipe.py` module docstring (lines 8-25). Notes that both Python CLI and Claude-side `arckit-build` skill interpret the same schema, and adds guidance for schema migrations.

### CI Gap

CI workflow now includes `pytest-asyncio` in the install step, so the 18 new tests will be visible to PR checks going forward. Note: the CI workflow still only runs `tests/codex/` and `tests/kimi/` — `tests/plugin/` (where the byo-llm suite lives) would need a separate workflow or path filter expansion to be fully gated.

tractorjuice:
Thanks for turning the blockers round so quickly. I verified all five on `ff5c3439` rather than reading the diff:

- `resolve_llm_config` now strips a trailing slash and a trailing `/v1`, so both URL conventions work
- the `ANTHROPIC_API_KEY` fallback is gone from `llm.py`
- config is written `0o600` under a `0o700` parent
- `pytest-asyncio` is declared in a `test` extra with `asyncio_mode = "auto"`
- `_is_sensitive_key` masks `api_key`/`token`/`secret` in `config get` and `list`

You also went past what I asked and fixed 6 and 7: `resolve_recipe_path` covers the nested `plugins/arckit-*/skills/*/recipes/` path now, and `all_searched` reports every candidate rather than only the directories that happen to exist. That is all good work and the code objections are closed.

Which leaves the architectural question, and I owe you a straight answer on it before you spend more time here. I went and researched it properly rather than asserting it.

## What I found

The premise the endpoint layer was built on has changed since you started. Claude Code speaks only the Anthropic Messages format, so historically pointing it at a local server meant running a translation proxy, which is a real reason to build the OpenAI-compatible path in Python instead. That is no longer true. Every serving stack your presets target now implements `/v1/messages` natively:

- vLLM, with a documented Claude Code integration page
- llama.cpp, PR #17570, January 2026
- Ollama, since v0.14.0
- SGLang, PR #18630
- LM Studio, January 2026

So for Ollama, SGLang, vLLM and OpenAI, the four presets in `local.py`, the configuration is two environment variables and the three `ANTHROPIC_DEFAULT_*_MODEL` remaps. Separately, `arckit init --ai codex` and `--ai opencode` already target hosts that speak the OpenAI protocol natively, so they point straight at the same servers with no translation anywhere in the path. That route needs nothing new at all and it is ArcKit's closest existing answer to BYO LLM.

That is not a criticism of your work. It is the platform moving underneath a design decision that was correct when you made it.

## Where that leaves the PR

I think this is two PRs that arrived as one, and they have different answers.

**The endpoint and wizard half**, roughly `local.py` plus the config and transport parts of `llm.py`, is superseded by the above. Maintaining ~1,100 lines of retry, config-precedence and health-check code to reach endpoints that the host CLIs already reach is a cost we would carry indefinitely for something two env vars do.

The genuinely missing piece was documentation. I grepped `docs/` and `README.md` for `ANTHROPIC_BASE_URL`, `base_url`, `Ollama`, `vLLM` and "local LLM" and got nothing, so the honest state of things was that ArcKit could do this and never told anyone. I have written that up in #805, covering all three routes, the per-server defects that are live right now (SGLang's `input_tokens: 0`, Ollama's `count_tokens` hang, vLLM's tool-parser flags), and the fact that Anthropic explicitly does not support routing Claude Code to non-Claude models. Please review it and tell me what I have got wrong, particularly on SGLang, since you have clearly run this against real hardware and I have not.

**The headless build harness**, `recipe.py` plus `state.py` plus the `build` command, is a different proposition and I do not think proxying replaces it. `/arckit:build` is Claude Code only and manual-invocation only, so there is no non-Claude runner today. A `pip install arckit && arckit build` path is a real gap and your DAG, wave and staleness work is sound. `--plan` produced a correct 9-wave ordering when I ran it.

If you want to pursue that half, two things need addressing first, and I would rather say them now than after you have rebased 85 commits:

1. **The tool set is too thin for what ArcKit asserts.** `TOOLS` is Read, Write and Glob. No Bash, no Grep, no web or MCP access, no subagents, and critically no hooks, so none of the governance layer runs: no provenance stamping, no `validate-arc-filename`, no secret detection, no citation traceability. The artefacts it produces will not pass ArcKit's own gates, which for a governance product is the thing that matters most.
2. **Output naming bypasses the ID helper.** `_build_user_prompt` hardcodes `projects/{proj}/ARC-001-{TYPE}-v1.0.md`. That gets multi-instance types wrong (ADR, DIAG, GRNT and the rest need `--next-num`) and ignores `SUBDIR_MAP`, so files land in the wrong place under the wrong names. `scripts/generate-document-id.mjs` is the single source of truth for this and should be shelling out or being ported. Related, `execute_target` returns success for any target whose output file merely exists, before the hash-based staleness logic in `state.py` gets a look in, which will mask exactly this class of bug.

## What I would suggest

Close this one, and open a fresh PR for the build harness alone against current `main`, without `local.py` and without the config wizard, with the two items above addressed. It will be a much smaller diff, it will review quickly, and it lands the part that is actually unique.

Entirely your call though, and if you would rather keep it as one PR and work through it here, I will review it that way. I did not want you rebasing and polishing a component I was going to argue against on grounds that have nothing to do with its quality.

tractorjuice:
Closing this one, so it is a decision rather than a PR that sits open indefinitely. To be clear about what that does and does not mean.

The BYO-LLM documentation gap you exposed is now closed: #805 is merged and live, covering all three routes, the per-server defects, and the governance consequences of swapping the model underneath the harness. That would not have been written without this PR, and the guide exists because you asked the right question.

The endpoint and wizard layer is the only part I am declining, and only because the serving stacks moved: vLLM, llama.cpp, Ollama, SGLang and LM Studio all serve `/v1/messages` natively now, so the translation problem it solves has gone away since you started.

**The build harness is still wanted.** `recipe.py`, `state.py` and the `build` command are the genuinely missing piece, there is no non-Claude runner today, and your DAG and wave ordering are correct. If you want to take it forward, a fresh PR against current `main` with just those three, plus the two things from my previous comment (wire the governance hooks, and route output naming through `generate-document-id.mjs` instead of hardcoding `ARC-001-{TYPE}-v1.0.md`), will review far faster than this one could. Reference this PR and I will pick it up promptly.

Thank you for the work here, and in particular for turning round all seven review findings as quickly as you did. None of this closure reflects on the quality of what you wrote.

Issue #723, opened 2026-07-30
doc-ids: port generate-document-id.sh to .mjs so it imports the registry instead of hardcoding a copy


tractorjuice:
## Summary

Port `generate-document-id.sh` to `generate-document-id.mjs` so it imports the doc-type registry instead of hardcoding a copy of it.

The motivation is **not** portability. It is that this script is the only bash helper that duplicates a registry, and that duplication has cost us a dedicated CI guard plus two shipped drift bugs.

## Why not portability

The obvious argument for `.mjs` is "node runs on more platforms than bash". That argument does not survive contact with the call sites:

- `generate-document-id.sh` is 1 of 6 scripts in `scripts/bash/`.
- **42 of the 66** commands that call it also call `create-project.sh` first. On a machine without bash those commands fail before they ever reach the ID generator.
- Converting one script buys nothing on a bash-less platform. The portability win only arrives if all six move, which is a much larger piece of work.
- Node is also not a uniform floor across the generated extensions. Codex hooks are `.mjs`; **Gemini hooks are `.py`** (`extensions/arckit-gemini/hooks/scripts/*.py`).

Node *is* already a hard dependency for Claude Code specifically: every entry in `plugins/arckit-claude/hooks/hooks.json` uses `command: "node"`, and there are already four non-hook `.mjs` scripts in `plugins/arckit-claude/scripts/` (`export-okf`, `import-okf`, `owm-to-mermaid`, `validate-handoff`). So the runtime is not the obstacle. It just is not the reason.

## The actual reason: the script duplicates the registry

`scripts/bash/generate-document-id.sh` line 85 carries its own copy of the registry:

```bash
MULTI_INSTANCE_TYPES="ADR DIAG DFD WARD DMC RSCH AWRS AZRS GCRS DSCT TNDR CMPT WGAM WCLM WVCH GOVR GCSR GLND GRNT CDAU"
```

in two tracked copies (`scripts/bash/` and `plugins/arckit-claude/scripts/bash/`). That is exactly why `scripts/check-multi-instance-parity.py` exists: it polices three registries, one of which is `config/doc-types.mjs` and two of which are these bash strings.

That guard was not preventive. The list drifted and shipped twice (TNDR/CMPT fixed in #566, `GRNT` found later while adding `CDAU`), and in #722 we deleted the script's own header comment for a third instance of the same class, five types behind the variable.

An `.mjs` implementation does:

```js
[...]

Issue #719, opened 2026-07-30
CI runs 52 of 1,445 pytest tests — the Python suites are ungated, so a failing test shows green


tractorjuice:
## Summary

CI runs pytest against two paths only:

```yaml
# .github/workflows/codex-plugin.yml
- name: Run Codex plugin tests
  run: python -m pytest tests/codex/test_codex_extension.py
- name: Test Kimi extension
  run: python -m pytest tests/kimi/
```

That is **52 of 1,445 collected pytest tests**. The other 1,393 are never executed by any workflow, so a Python test can fail indefinitely while every check shows green.

| Directory | pytest tests | Run in CI |
|---|---|---|
| `tests/plugin/` (11 `.py` files) | 1,321 | **no** |
| `tests/vibe/` | 38 | **no** |
| `tests/codex/` | 35 | yes |
| `tests/kimi/` | 17 | yes |
| `tests/paperclip/` | 11 | **no** |
| `tests/gemini/` | 9 | **no** |
| `tests/copilot/` | 7 | **no** |
| `tests/opencode/` | 7 | **no** |

**Important scope note:** the JavaScript side *is* covered. `lint-markdown.yml` runs `node --test tests/plugin/*.test.mjs tests/plugin/test_hook_utils.mjs` plus four handoff-schema validators, so the 15 `.test.mjs` hook suites in `tests/plugin/` are gated. This issue is strictly about the **Python** suites, including the 11 `.py` files that sit in the same directory as those covered `.mjs` files.

## Two confirmed consequences

**1. #711 ships a test suite that has never run, with CI green.** `tests/plugin/test_byo_llm.py` adds 307 lines and 5 async tests covering LLM retry and backoff. All five fail:

```
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_connect_error
FAILED tests/plugin/test_byo_llm.py::test_call_exhausts_retries
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_429
FAILED tests/plugin/test_byo_llm.py::test_call_no_retry_on_500
FAILED tests/plugin/test_byo_llm.py::test_call_retries_on_timeout
```

with `async def functions are not natively supported` and `PytestUnknownMarkWarning: Unknown pytest.mark.asyncio`. `pytest-asyncio` is neither declared nor installed. Both checks on that PR pass. Without a local run there is no signal at all that the feature's most safety-relevant tests are inert.

[...]

PR #726, opened 2026-07-30
fix(pages): derive project buckets from SUBDIR_MAP so /arckit:pages stops crashing

tractorjuice:
`/arckit:pages` has been unable to rebuild the dashboard since v6.7.0, in any repo containing an `audits/` or `framework/` directory. Found while verifying that the document-map pipeline still worked after #723 and #725; it is pre-existing and reproduces on `origin/main`.

## The crash

`scanProject()` in `sync-guides.mjs` routes each artefact into `project[<camelCased SUBDIR_MAP directory>]`, **deriving** those keys from `SUBDIR_MAP`. The bucket list it pushes into was **hand-written**. Registering a doc-type with a new subdirectory therefore left an undefined array:

```
TypeError: Cannot read properties of undefined (reading 'push')
    at scanProject (sync-guides.mjs:374)
```

It shipped twice and was never reported, because the crash only fires in a repo that actually has the directory:

- `audits` arrived with `CDAU` in v6.7.0
- `framework` arrived with `FWRK` in #714

This repo's own `projects/` has an `audits/` directory, which is how it surfaced. It is also why `docs/manifest.json` is dated 27 July: the hook that regenerates it has been dying first.

## The fix

Buckets are derived from `SUBDIR_MAP` by the same `subdirKey()` helper the routing loop uses, so initialisation and routing cannot diverge again. Same principle as #723: stop maintaining a second copy of something the registry already knows.

`tests/plugin/sync-guides-buckets.test.mjs` asserts the **derivation** rather than the resulting names, so it cannot pass while silently omitting a directory. It fails 2 of 4 against the previous shape.

## Eight Node suites, 64 tests, were never run by CI

`lint-markdown.yml` globbed `tests/plugin/*.test.mjs` and listed exactly one `test_*.mjs` file explicitly. The other eight were unrun:

`test_update_manifest.mjs` (the `docs/manifest.json` coverage), `test_graph_inject.mjs`, `test_graph_utils.mjs`, `test_export_okf.mjs`, `test_import_okf.mjs`, `test_okf_frontmatter.mjs`, `test_provenance_okf_frontmatter.mjs`, `test_kimi_hook_adapter.mjs`.

The step's own comment claimed the glob "avoids the per-file enumeration drift that left suites unrun". It did not, for the second naming convention. Both patterns are now globbed, taking the job from **166 tests to 266**. Same class of gap as #719, where pytest ran 52 of 1,445.

All eight pass unchanged, so nothing was hiding behind the gap.

## Reported, not fixed here

`sync-guides.mjs` rebuilds `docs/manifest.json` and `docs/index.html` by scanning the **working tree**, including the gitignored `projects/`. Consequences observed while testing:

- the committed manifest holds **33 projects**; regenerating locally replaced them with the **2** on this machine's disk
- the regenerated `index.html` dropped the hand-added Google Analytics tag (`G-SX6P1GM31K`), so the committed file has diverged from what `pages-template.html` produces

Both were reverted rather than committed. This also sits oddly against #690, which taught `update-manifest.mjs` **not** to index gitignored artefacts — the two writers of the same file now disagree about whether to filter. Worth its own issue; I have not filed one.

## Verified

- hook exits 0 and populates the previously-crashing `audits` bucket
- 266 node tests pass (was 166), 1236 pytest passed / 225 skipped
- markdownlint clean

https://claude.ai/code/session_01Fn6shRoNVcztoUYrDe7cvT


tractorjuice:
Filed the manifest/index.html provenance hazard noted above as #727. It turned out sharper than described here: the committed `docs/manifest.json` is the output of the generator #691 reverted, carrying 4 of the 12 keys `sync-guides.mjs` produces, and it cannot be safely regenerated until that hook gets the `git check-ignore` guard #690 added to `update-manifest.mjs`.

This PR is a prerequisite either way: `/arckit:pages` currently crashes before it can regenerate anything.

PR #728, opened 2026-07-30
fix(pages): guard the published index against gitignored files, preserve hand-added head content (#727)

### LINKED ISSUE #727: docs/manifest.json is the reverted #683 generator's output: 4 keys where the site expects 12, and regenerating it publishes local gitignored projects
## Summary

The committed `docs/manifest.json` is the output of the `generate-docs-manifest.py` script that #691 **reverted**. The revert removed the generator and left its output in place, so the published manifest carries 4 of the 12 keys the real generator produces, and nothing has regenerated it since.

It cannot simply be regenerated, because the real generator has no gitignore filter and ArcKit gitignores its own `projects/`. Running `/arckit:pages` in this repo replaces the published index with a scan of whatever happens to be on the local disk.

Found while verifying the document-map pipeline after #723 and #725.

## The schema gap

```
COMMITTED   (4): generated, global, projects, repository

REGENERATED (12): defaultDocument, dependencyGraph, generated, global,
                  guideCategoryOrder, guideSectionOrder, guides, projectHealth,
                  projects, repository, roleGuides, typeCategories
```

`plugins/arckit-claude/templates/pages-template.html` references `manifest.json` three times. The published copy is missing `guides`, `roleGuides`, `guideSectionOrder`, `guideCategoryOrder`, `typeCategories`, `defaultDocument`, `dependencyGraph` and `projectHealth`. I have not characterised exactly how the rendered site degrades, only that eight keys it is built against are absent.

## How it got here

- **22c86e76** (#683) `ci: generate docs/manifest.json from disk instead of by hand` committed a manifest produced by `scripts/generate-docs-manifest.py`, whose `projects[]` entries are *guide groups* (`guides-architecture`, `guides-compliance`, …) rather than ArcKit projects.
- **12ca1cc1** (#690) touched the same file.
- **cc9cf7ae** (#691) reverted the generator, correctly: *"That was a duplicate, inferior implementation of something the product already does. sync-guides.mjs, driven by /arckit:pages, is the real generator. It builds a far richer schema."*

The reasoning was right; the cleanup was incomplete. The generator went, the inferior output stayed, and `sync-guides.mjs` has not run against this repo since.

## Why regenerating is not a one-liner

`sync-guides.mjs` scans the working tree with **no git filtering at all** (`grep -c "ls-files\|isTracked"` returns 0). ArcKit gitignores its own `projects/`. Running the hook here produced:

```
projects: 33 -> 2
```

The 33 guide-group entries were replaced by the two real project directories on my machine, neither of which is tracked. Anyone running `/arckit:pages` in a checkout of this repo would publish their local project artefacts to arckit.org.

This is exactly the failure #690 fixed **in the other writer**. `update-manifest.mjs` gained a `git check-ignore -q` guard so that "a published index must not reference a file that will never be published". `sync-guides.mjs`, which rewrites the whole file rather than appending to it, never got one.

## Stale ownership comment

`plugins/arckit-claude/hooks/update-manifest.mjs` still documents the arrangement as:

```
 * Ownership of docs/manifest.json:
 *   - THIS hook owns project artefact entries (projects[] groups) …
 *   - scripts/generate-docs-manifest.py owns the guide/template/article/global
 *     index in the ArcKit repo itself, rebuilt from git-tracked sources.
 *   The two do not overlap once the gitignore guard is in place, because
 *   ArcKit gitignores its own projects/.
```

`scripts/generate-docs-manifest.py` was deleted in cc9cf7ae. The actual co-owner is `sync-guides.mjs`, and the non-overlap claim does not hold: `sync-guides.mjs` writes the entire file, `projects[]` included, with no gitignore guard. The comment describes an arrangement that has not existed since #691.

## Second casualty: the analytics tag

`sync-guides.mjs` also regenerates `docs/index.html` from `pages-template.html`. The committed file has diverged from the template:

```
docs/index.html      : 2 occurrences of G-SX6P1GM31K
pages-template.html  : 0
```

So the Google Analytics tag was added to the published page by hand, and **any** `/arckit:pages` run silently removes it. Regenerating the manifest and regenerating `index.html` are the same action, so the two problems have to be solved together.

I reverted both files rather than committing them while investigating.

## Suggested resolution

1. Give `sync-guides.mjs` the same `git check-ignore` guard `update-manifest.mjs` got in #690, so a published index can never reference an unpublishable file. That makes regeneration safe in any repo, not just this one.
2. Restore the analytics tag to `pages-template.html` so the generated page matches what is published, or move it to a fragment the generator preserves.
3. Regenerate `docs/manifest.json` with the real generator so the published file carries the 12-key schema the template is built against.
4. Correct the ownership comment in `update-manifest.mjs` to name `sync-guides.mjs` and drop the non-overlap claim.

Worth deciding explicitly whether `docs/manifest.json` should be a committed artefact at all, given it is machine-generated from tracked sources and has now been stale-by-construction twice.

## Context

Follow-up to #691 (the revert), #690 (the gitignore guard added to the other writer) and #683 (the reverted generator). Adjacent to #726, which fixes a separate crash in the same hook that has been preventing `/arckit:pages` from running at all since v6.7.0 — that crash is why this went unnoticed.

tractorjuice:
Closes #727.

## Only one of the two `docs/manifest.json` writers had a gitignore guard

#690 taught `update-manifest.mjs` that *"a published index must not reference a file that will never be published"*. `sync-guides.mjs`, which rewrites the **whole** manifest rather than appending one entry, never got the same rule. ArcKit gitignores its own `projects/`, so running `/arckit:pages` here replaced the published index with a scan of whatever was on that machine's disk.

`sync-guides.mjs` now applies the same rule, batched into a single `git check-ignore --stdin` rather than one spawn per path, because it walks hundreds of entries under a 15s hook budget. It fails open when git is unavailable, still indexes untracked-but-not-ignored artefacts (the normal case for one written moments ago), and reports what it omitted instead of dropping entries silently.

Worth knowing for the tests: `git check-ignore` deliberately does **not** report a *tracked* file as ignored, because a tracked file is published whatever the rules say. So the guard can only fire on untracked-and-ignored — exactly ArcKit's own situation. The fixture builds both cases, and my first version of the test was wrong for precisely this reason.

## Hand-added `<head>` content was destroyed on every run

`docs/index.html` is generated wholesale from `pages-template.html`, so arckit.org silently lost its Google Analytics tag.

The tag cannot simply go in the template. **The template ships to every ArcKit user**, and baking one site's measurement ID into it would put that ID on everybody's generated page. So instead, content between `<!-- ARCKIT:PRESERVE -->` and `<!-- /ARCKIT:PRESERVE -->` is carried across from the existing page into the regenerated one — analytics tags, verification meta, custom fonts. Both template copies ship the markers, and a test asserts neither contains anything shaped like a measurement ID.

## `docs/manifest.json` rebuilt with the real generator

It was the output of the generator **#691 reverted**. #683 added `scripts/generate-docs-manifest.py`; #691 removed it as *"a duplicate, inferior implementation"* — correctly — but left its output committed, and the crash fixed in #726 meant nothing could regenerate it since.

| | before | after |
|---|---|---|
| keys | 4 | 12 |
| guides | — | 204 |
| role guides | — | 18 |
| projects | 33 pseudo-projects (really guide groups) | 0 |

`projects: []` is the honest answer for a repo that gitignores its own `projects/`.

## `docs/index.html` is deliberately NOT regenerated

The published page is hand-maintained and has diverged from `pages-template.html`. Regenerating it drops `<title>`, the favicons, `theme.css`/`heroes.css`, the GDS font-override block and the custom component styles. It also **does not reference `manifest.json` at all**, so it is not a consumer of the file this PR fixes.

The only change to it is wrapping the existing analytics block in the preserve markers, so the next regeneration keeps it. `+2 lines`. The template/published-page divergence needs its own issue — I have not filed one yet.

## Ownership comment

`update-manifest.mjs` still described an arrangement that ended in #691: it named the deleted `generate-docs-manifest.py` and claimed the two writers *"do not overlap once the gitignore guard is in place"*. They did overlap, and the guard was on only one of them. Corrected to name `sync-guides.mjs` as the full generator and this hook as the incremental updater, both now guarded.

## Tests

`tests/plugin/sync-guides-publish-guard.test.mjs` — 8 tests, end to end against real temporary git repos, since the guard's whole point is agreeing with what git actually ignores:

- tracked artefact **is** indexed
- gitignored artefact is **not**
- untracked-but-not-ignored artefact **is** (guards against regressing into "only committed files count")
- omission is reported, not silent
- fails open outside a git repo
- content inside the markers survives regeneration
- content outside the markers does not
- neither template copy carries a measurement ID

Verified: `1241 passed, 225 skipped` (pytest), `274 passed` (node), all 10 repo guards pass, markdownlint clean across 3535 files.

https://claude.ai/code/session_01Fn6shRoNVcztoUYrDe7cvT


tractorjuice:
Split the template/published-page divergence out as #729 — it blocked the obvious "just regenerate `docs/index.html`" fix here, and it turns out the two `pages-template.html` copies have drifted from each other too (47 lines), with the `.arckit` one winning at runtime.

tractorjuice:
Reworked after review. The original approach was wrong in two ways, both found by testing against real repo shapes rather than reasoning about them:

**1. It missed `dependencyGraph` entirely.** That comes from a separate `scanAllArtifacts()` call and keys its nodes by document ID rather than storing an array, so a generic deep walk that splices arrays left **7 gitignored paths and 20 edges** published — the exact leak the guard exists to close.

**2. It was too broad.** Walking the whole manifest meant a repo that gitignores its generated `docs/` lost all **204 guides and 18 role guides** from the dashboard. Silently emptying the site is worse than the leak.

The guard now filters where project directories are discovered, so `projects[]`, `global[]`, `dependencyGraph` and `projectHealth` are correct by construction. Graph nodes carry their project directory name; edges drop when either endpoint goes, so nothing dangles.

**Checked against the real test repos**: most carry no `.gitignore` at all, and the two that do (`v7-nhs-appointment`, `v49-ca-overlay`) have no `docs/` or `projects/` rules — so this is a no-op for every one of them.

Three of the 11 tests exist specifically because the first implementation failed them. Node suite is 277, up from 274.

PR #733, opened 2026-08-02
docs(mcp): add a keyless web search example

georgeatparallel:
ArcKit's optional MCP section currently points readers to vector search, but its research workflows can also benefit from current web results. This adds a project-local, keyless remote example for live search and URL fetching without changing ArcKit's bundled servers or agent defaults.

## What changed

- Added a `parallel-search` `.mcp.json` example using the existing remote HTTP configuration shape
- Documented the `web_search` and `web_fetch` tools
- Kept the canonical and plugin guide copies in sync and added an Unreleased changelog entry

## Validation

- `python3 scripts/check-guide-parity.py --check`
- Markdown lint on all three changed files
- `python3 scripts/check_references.py` (791 Markdown files, no broken references)
- `git diff --check`

Disclosure: I work at Parallel.ai, which operates this MCP endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Optional Parallel Search MCP setup.** The MCP guide now includes a
+  project-local, keyless remote configuration for live web search and URL
+  fetching. It does not change ArcKit's bundled MCP servers or agent defaults.
+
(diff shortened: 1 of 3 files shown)


tractorjuice:
Thanks for the PR, and genuinely, thanks for the up-front disclosure and for getting the mechanics right. Keeping both guide copies in sync and running `check-guide-parity.py` is more care than most drive-by contributions here show.

I'm going to close this one, for three reasons.

Claude Code already provides `WebSearch` and `WebFetch` as built-in tools, and ArcKit's research agents (`arckit-research`, `arckit-datascout`, `arckit-grants`) already allowlist them. A third-party server covering the same ground would mean adding a dependency to replace something readers already have.

ArcKit's primary audience is UK public sector work classified up to OFFICIAL-SENSITIVE. Documenting an unauthenticated remote endpoint means every query a reader runs through it leaves their environment for a third party they have no contract with. That is a decision each organisation should make deliberately through its own assurance process, not one ArcKit should nudge them into from a setup guide.

The CHANGELOG entry is also the wrong shape for this repo. The Unreleased section tracks changes to ArcKit itself rather than additions to the third-party list.

None of that is a judgement on Parallel Search as a product. If ArcKit users start asking for it, the third-party MCP list is the right place for it to land.

PR #740, opened 2026-08-07
feat(eu): add /arckit:eu-cloud-sovereignty

umag:
Adds `/arckit:eu-cloud-sovereignty` (`EUCSF`) to the existing EU overlay. Companion to #739, but **independent** — both branch from `main` and touch different plugins, so they can merge in either order.

## Why here rather than in a national overlay

The European Commission published the **Cloud Sovereignty Framework** (v1.2.1, October 2025) to supplement security assurance with sovereignty-specific safeguards in cloud procurement. It applies across member states, and it is starting to be adopted as a national yardstick rather than only a Commission procurement tool — so it seemed to belong in `arckit-eu` rather than buried in one country's overlay, even though the Netherlands is currently the clearest adopter.

## What it does

Assesses a cloud service against the eight Sovereignty Objectives using the framework's own weights:

| Objective | Weight |
|---|---|
| SOV-1 Strategic Sovereignty | 15% |
| SOV-2 Legal & Jurisdictional Sovereignty | 10% |
| SOV-3 Data & AI Sovereignty | 10% |
| SOV-4 Operational Sovereignty | 15% |
| SOV-5 Supply Chain Sovereignty | 20% |
| SOV-6 Technology Sovereignty | 15% |
| SOV-7 Security & Compliance Sovereignty | 10% |
| SOV-8 Environmental Sustainability | 5% |

It records a Sovereignty Effectiveness Assurance Level per objective from SEAL-0 (no sovereignty) to SEAL-4 (full digital sovereignty), computes the weighted Sovereignty Score the framework defines as an award criterion, and checks each objective against its required minimum SEAL.

## Two things the command is deliberately loud about

Both are easy to get wrong, and both change the outcome of a procurement:

- **Minimum SEAL levels come from the tender specification, not from the framework.** The framework supplies the scale; the contracting authority supplies the floor. Reading the framework as prescribing a floor is the single most common misreading.
- **A supplier's self-declared SEAL is an unverified claim until the assessor records evidence.** The template carries an **evidence column per objective**, drawn from the framework's own contributing factors — which are observable things rather than assertions: who holds decisive authority over the service, which legal system governs the contract, whether the customer alone holds cryptographic access, where support staff sit and under whose jurisdiction, the provenance of hardware, firmware and software including sub-suppliers, and whether APIs and licences actually permit exit.

[...]

PR #795, opened 2026-08-14
fix(eu): correct three SOV weights and the SEAL-3 name against the framework

umag:
Three of the eight Sovereignty Objective weights, and the name of SEAL-3, contradict the framework this command implements. Both shipped in #740 — mine to fix.

Sources: [Implementation guidance](https://commission.europa.eu/document/download/2ad80a48-166f-4c77-a513-80c53ca2a128_en?filename=Cloud+Sovereignty+Framework+-+Implementation+guidance.pdf) p.7 and the [Annex — Sovereignty assessment calculator](https://commission.europa.eu/document/download/3acb8fe8-8a4a-4339-ae74-f56138d913d1_en?filename=Annex+-+Sovereignty+assessment+calculator.xlsx) cells `D4`–`D231`. Both agree.

## The weights

| Objective | Shipped | Framework | |
|---|---|---|---|
| SOV-1 Strategic | 15% | **20%** | ✗ |
| SOV-2 Legal & Jurisdictional | 10% | 10% | ✓ |
| SOV-3 Data & AI | 10% | 10% | ✓ |
| SOV-4 Operational | 15% | 15% | ✓ |
| SOV-5 Supply Chain | 20% | **10%** | ✗ |
| SOV-6 Technology | 15% | 15% | ✓ |
| SOV-7 Security & Compliance | 10% | **15%** | ✗ |
| SOV-8 Environmental | 5% | 5% | ✓ |

## Why it survived two independent verifications

**The defect is a permutation.** The wrong set also sums to 100.

My PR body on #740 said "weights verified programmatically to sum to 100". You independently confirmed "15 + 10 + 10 + 15 + 20 + 15 + 10 + 5 = 100". Both statements were true. Both were invariant to the error.

So the guard added here asserts each weight **individually**, across canonical files *and* generated mirrors — and it was run on unfixed `main` **before** the correction:

```
[FAIL] plugins/arckit-eu/commands/eu-cloud-sovereignty.md:198: SOV-1 weight is 15%, the framework says 20%
[FAIL] plugins/arckit-eu/commands/eu-cloud-sovereignty.md:202: SOV-5 weight is 20%, the framework says 10%
[FAIL] plugins/arckit-eu/commands/eu-cloud-sovereignty.md:204: SOV-7 weight is 10%, the framework says 15%
… 48 [FAIL] lines across 7 files, all naming exactly those three objectives
```

A guard written after the fix has never been observed to fail. The existing sum-to-100 assertion stays, but demoted — it passed throughout the defect's life.

`scripts/tests/test-csf-weights.mjs` is wired into `lint-markdown.yml` with its own `run:` step, beside the three existing mjs guards. A guard with only a path filter never executes.

## SEAL-3 is "Technological Sovereignty"

Guidance p.2–3, and again in the p.10 table. We shipped **"Digital Resilience"** — inside the block instructing *"do not paraphrase or abbreviate"*. SEAL-0, SEAL-2 and SEAL-4 wording had also drifted (`governed entirely in` → `by`; `EU law applicable and enforceable` → `EU jurisdictions apply`; `subject only to EU law` → `EU jurisdiction`). All five rows now match p.2–3 verbatim, in both the command's definition block and the template's own SEAL table — correcting only the command would have left the two disagreeing about the name of the level being assessed, which is worse than being uniformly wrong.

p.2–3 is treated as normative for the definition block; p.10 gives a second, non-identical rendering and is the per-objective requirements view.

## Two things the correction dragged with it

**The template carries the weights in 12 places, not 2.** Three tables (Executive Summary, Objective Weights, Scored Result) plus eight inline `(Weight: NN%)` subsection headers. Fixing a subset would have left the document contradicting itself.

**A derived claim in the guide went stale.** It read *"Supply chain carries the heaviest weight, and technology, strategic and operational sovereignty together carry 45%"* — both halves false once corrected: Strategic Sovereignty is heaviest at 20%, and the three-objective sum is 50%. Corrected in your voice rather than left to contradict the table above it. That sentence traces back to my framing in #740, built on the wrong number.

## Verification

```
test-csf-weights.mjs               PASS (128 occurrences, 7 files)
4 mjs guards                       all pass
```

`check-guide-parity.py --sync` was needed after editing `docs/guides/` — the converter copies plugin→extensions, not root→plugin, so the mirror drifts otherwise.

## Merge order

Companion to the overall-SEAL and scoring PRs. All three touch the command, template, checklist and CHANGELOG. This one should land **first** — its weight values are load-bearing for the scoring PR's data file.

### DIFF
--- a/.arckit/templates/eu-cloud-sovereignty-template.md
+++ b/.arckit/templates/eu-cloud-sovereignty-template.md
@@ -77,11 +77,11 @@
 
 | Level | Name | Definition |
 |-------|------|------------|
-| SEAL-0 | No Sovereignty | Service, technology or operations under exclusive control of non-EU third parties, governed entirely in non-EU jurisdictions |
+| SEAL-0 | No Sovereignty | Service, technology or operations under exclusive control of non-EU third parties, governed entirely by non-EU jurisdictions |
 | SEAL-1 | Jurisdictional Sovereignty | EU law formally applies with limited practical enforceability; service, technology or operations under exclusive control of non-EU third parties |
-| SEAL-2 | Data Sovereignty | EU law applicable and enforceable, with material non-EU dependencies remaining; under indirect control of non-EU third parties |
-| SEAL-3 | Digital Resilience | EU law applicable and enforceable, EU actors exercising meaningful but not full influence; under marginal control of non-EU third parties |
-| SEAL-4 | Full Digital Sovereignty | Technology and operations under complete EU control, subject only to EU law, with no critical non-EU dependencies |
+| SEAL-2 | Data Sovereignty | EU jurisdictions apply, with material dependencies remaining; service, technology or operations under indirect control of non-EU third parties |
+| SEAL-3 | Technological Sovereignty | EU jurisdictions apply, EU actors exercising meaningful but not full influence; service, technology or operations under marginal control of non-EU third parties |
+| SEAL-4 | Full Digital Sovereignty | Technology and operations under complete EU control, subject only to EU jurisdiction, with no critical non-EU dependencies |
 
 **Formula**: Sovereignty Score = Σ over the eight objectives of (Score(SOVn) / Max.Score(SOVn)) × Weight(SOVn), expressed as a percentage.
 
--- a/.arckit/templates/eu-cloud-sovereignty-template.md
+++ b/.arckit/templates/eu-cloud-sovereignty-template.md
@@ -27,13 +27,13 @@
 
 | Objective | Weight | SEAL Claimed | SEAL Evidenced | Minimum SEAL (tender) | Meets Minimum? |
 |-----------|--------|-------------|-----------------|------------------------|-----------------|
-| SOV-1 Strategic Sovereignty | 15% | [SEAL-0 to SEAL-4] | [SEAL-0 to SEAL-4] | [SEAL / Not yet set] | ☐ |
+| SOV-1 Strategic Sovereignty | 20% | [SEAL-0 to SEAL-4] | [SEAL-0 to SEAL-4] | [SEAL / Not yet set] | ☐ |
 | SOV-2 Legal & Jurisdictional Sovereignty | 10% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
 | SOV-3 Data & AI Sovereignty | 10% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
 | SOV-4 Operational Sovereignty | 15% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
-| SOV-5 Supply Chain Sovereignty | 20% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
+| SOV-5 Supply Chain Sovereignty | 10% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
 | SOV-6 Technology Sovereignty | 15% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
-| SOV-7 Security & Compliance Sovereignty | 10% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
+| SOV-7 Security & Compliance Sovereignty | 15% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
 | SOV-8 Environmental Sustainability | 5% | [SEAL] | [SEAL] | [SEAL / Not yet set] | ☐ |
 
 > ⚠️ This document is an **assessment record**, not a certification. There is no published EU list of providers assessed against this framework — no commercial cloud provider is named as sovereign, compliant, or achieving any SEAL level in this document.
@@ -89,13 +89,13 @@
 
 | Objective | Weight |
 |-----------|--------|
-| SOV-1 Strategic Sovereignty | 15% |
+| SOV-1 Strategic Sovereignty | 20% |
 | SOV-2 Legal & Jurisdictional Sovereignty | 10% |
 | SOV-3 Data & AI Sovereignty | 10% |
 | SOV-4 Operational Sovereignty | 15% |
-| SOV-5 Supply Chain Sovereignty | 20% |
+| SOV-5 Supply Chain Sovereignty | 10% |
 | SOV-6 Technology Sovereignty | 15% |
-| SOV-7 Security & Compliance Sovereignty | 10% |
+| SOV-7 Security & Compliance Sovereignty | 15% |
 | SOV-8 Environmental Sustainability | 5% |
 | **Total** | **100%** |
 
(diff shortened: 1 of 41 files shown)


tractorjuice:
## Code Review

**Verdict:** Correct. I verified the weights and the SEAL wording independently against both primary sources rather than against the PR body, and every claim holds. Land this one first, as you propose. One minor gap in the new guard, plus a cross-PR merge hazard noted at the end.

### Verified against the primary sources

I downloaded both Commission artefacts and checked them directly.

**Weights.** Implementation guidance p.7 gives, in its own table, `Weight: 20% / 10% / 10% / 15% / 10% / 15% / 15% / 5%` for SOV-1 to SOV-8. The Annex calculator agrees at the cells you cite:

```
D4=0.2   D45=0.1  D76=0.1   D102=0.15
D133=0.1 D169=0.15 D195=0.15 D231=0.05
```

So SOV-1 20%, SOV-5 10%, SOV-7 15% are right and `main` is wrong on exactly those three. The permutation explanation is correct: the shipped set is the same eight values misassigned, which is why every sum check passed.

**SEAL wording.** Guidance p.2 to p.3 confirms all five rows verbatim, including the two things a reviewer might otherwise flag as errors in this PR:

- SEAL-3 is `Technological Sovereignty`. "Digital Resilience" appears nowhere in the framework.
- SEAL-2 really does read "EU jurisdictions apply, with material dependencies remain". The source's grammar is broken there; keeping it verbatim is the right call given the surrounding "do not paraphrase" instruction. Worth a one-line note in the file so a future contributor does not "fix" it.

`governed entirely by`, `EU jurisdictions apply` and `subject only to EU jurisdiction` are all confirmed against the same table.

**Derived prose.** The rewritten guide sentence is arithmetically right: 20% is the unique maximum, and SOV-1 + SOV-6 + SOV-4 = 50%.

**Guard.** `test-csf-weights.mjs` is well built, and the design point is the right one: assert each objective individually, glob the mirrors, and wire it into `lint-markdown.yml` with its own `run:` step. It passes here (128 occurrences, 7 files), and it correctly caught a weight regression I introduced by hand while testing the merge below.

### MINOR: the new checklist item is not covered by the new guard

The PR replaces the sum-only checklist item with an explicit eight-weight list, which is the right fix. But that list is prose, and the guard matches only table rows and `(Weight: NN%)` headers, so the list itself is unguarded.

Reproduced on this branch by putting the original wrong values back into the checklist prose only:

```
SOV-1 Strategic Sovereignty 20%  ->  15%
SOV-5 Supply Chain Sovereignty 10%  ->  20%

node scripts/tests/test-csf-weights.mjs
[PASS] 128 SOV weight occurrence(s) across 7 file(s) all match the framework
```

`sync-shared-assets.py --check` caught it, but only because I edited one of the 20 copies. A wrong value in the canonical core copy would sync cleanly to all 20 and the guard would still pass, which is the same shape of blind spot the PR is written to close.

Suggested fix: add `plugins/arckit-claude/references` to `ROOTS` and teach the matcher a third shape, `SOV-<n> <Name> <NN>%` in prose.

### Cross-PR: the merge presents the pre-fix text as a merge side

Flagging here because this PR owns the corrected text. #796 and #797 both branch from `f1784387`, so each conflicts with this one on the SEAL definition block and on the weight-bearing template headers. Merging all three locally, the conflict looks like this:

```
<<<<<<< HEAD
  - **SEAL-3 Technological Sovereignty**: EU jurisdictions apply, ...
=======
  - **SEAL-3 Digital Resilience**: EU law applicable and enforceable, ...
>>>>>>> pr-796
```

A resolution that leaves both lines in place ships the command with two contradictory SEAL-3 names six lines apart, inside the block whose own instruction says "do not paraphrase or abbreviate". I tried it: `test-csf-weights.mjs` passes and `markdownlint-cli2` reports 0 issues. Weight regressions are caught by the new guard, SEAL level names are caught by nothing.

Two suggestions:
1. Have #796 and #797 rebased on `main` after this lands, rather than resolving at merge time. Their bodies predict a CHANGELOG rebase; it is actually eight files including `.github/workflows/lint-markdown.yml` and `docs/llms.txt`.
2. Consider extending the guard, here or in a follow-up, to assert the five SEAL level names. It is the same defect class this PR exists to prevent, and it is currently one careless conflict resolution away.

### POSITIVE

- Running the guard red on unfixed `main` before writing the fix, and quoting the failure, is the right way to demonstrate a guard actually bites. Worth keeping as the house pattern.
- Correcting SEAL-2 and SEAL-4 alongside SEAL-3, rather than only the headline defect, avoids leaving the command and template disagreeing about the level being assessed.
- Catching that the guide's "supply chain carries the heaviest weight" sentence was derived from the wrong number is the kind of second-order breakage that usually survives this sort of fix.

---

Review verified against: Implementation guidance PDF (13pp) and Annex calculator XLSX, both fetched from the Commission URLs in the PR body.

tractorjuice:
Rebased onto `main` (v6.10.0 had consumed the `[Unreleased]` section the entry was written under — the only conflict was `CHANGELOG.md`, re-homed under a fresh `[Unreleased]`), and addressed the review.

**MINOR — the new checklist item was not covered by the new guard.** Confirmed: reverting `SOV-1` and `SOV-5` in the checklist prose alone still ran green. The list is prose rather than a table row or a `(Weight: ...)` header, and its 31 copies sit outside the guarded roots.

`test-csf-weights.mjs` now recognises a third shape, `SOV-<n> <objective name> <NN>%`, and collects `quality-checklist.md` by basename so a new plugin is covered when it is added rather than when someone remembers to list it. Coverage: **128 occurrences over 7 files → 376 over 38**.

Anchored on the objective's own name rather than a loose `SOV-<n>...<NN>%` scan, for two reasons. The loose form false-positives on that same checklist line — "weights summing to 100%" reads as SOV-8 being 100% — which is the case the guard's own header comment already warned about. And matching the name buys a second assertion: a code paired with **another objective's name** now fails too, which is the permutation defect wearing its other face, and the one that let `SEAL-3` ship as "Digital Resilience".

Both red runs verified:

```
[FAIL] plugins/arckit-eu/references/quality-checklist.md:468: SOV-1 weight is 15%, the framework says 20%
[FAIL] plugins/arckit-eu/references/quality-checklist.md:468: SOV-5 weight is 20%, the framework says 10%

[FAIL] plugins/arckit-eu/references/quality-checklist.md:468: SOV-1 is named "Technology Sovereignty", the framework says "Strategic Sovereignty"
```

**On the SEAL-2 grammar** — agreed, and marked in the command as verbatim-on-purpose with the reason, so it is not silently repaired out of step with the normative text a tender is assessed against.

**On the cross-PR merge hazard** — handled on the other two branches rather than here: #796 and #797 now carry these same corrections, so both sides of every shared line make the identical change. Git resolves them with no conflict, and no merge order can reintroduce the wrong values. Verified by diffing every SEAL definition and SOV weight row across all three branches.

Green: full CI guard set, 1541 pytest, markdownlint over 3800 files. No conflicts with `main`.

tractorjuice:
Merged as 7afb90a8 — first of the three, as you proposed.

`test-csf-weights.mjs` is live in `lint-markdown.yml` on `main` and passing at 376 weight occurrences across 38 files.

Also added a note to #740 recording that the weight table in its description is wrong as merged, with the correction and why a permutation slipped past both the review and the sum-only checklist item. Anyone reading that PR later would otherwise take the table at face value.

PR #796, opened 2026-08-14
fix(eu): add the framework's overall-SEAL rule to eu-cloud-sovereignty

umag:
`/arckit:eu-cloud-sovereignty` records a SEAL per objective and frames "minimum SEAL" as a rejection gate — but has no aggregate SEAL at all, so it cannot express the gate it exists to evaluate.

The framework defines one. Source: [Implementation guidance](https://commission.europa.eu/document/download/2ad80a48-166f-4c77-a513-80c53ca2a128_en?filename=Cloud+Sovereignty+Framework+-+Implementation+guidance.pdf) and the [Annex calculator](https://commission.europa.eu/document/download/3acb8fe8-8a4a-4339-ae74-f56138d913d1_en?filename=Annex+-+Sovereignty+assessment+calculator.xlsx).

## It is a minimum, never an average

Guidance p.9:

> The overall SEAL level is **the lowest SEAL level achieved in any of the objectives**.

The calculator implements it at finer granularity — `F2 = "SEAL-"&MIN(H5:H251)`, a minimum over every *answered criterion*, each answer carrying its own SEAL in column `F`.

These agree only because an objective's SEAL is itself the minimum across its criteria, so both levels are stated and cited separately rather than conflated:

- **Per-objective SEAL** = minimum across that objective's criteria (calculator `F2`)
- **Overall SEAL** = minimum across the eight objectives (guidance p.9)

A minimum-of-minimums. A service at SEAL-4 on seven objectives and SEAL-1 on one is **SEAL-1**. An assessor averaging or taking the modal SEAL reaches the opposite admission decision, so the command says so explicitly and the quality checklist rejects averaging and modal language.

The artefact now records the Overall SEAL **and its governing criterion** — which objective, and where identifiable which contributing factor, set the minimum.

## Three supporting items from the same source

**The inert-criterion property** (p.9): *"When all responses grant SEAL-4, it must be understood that the criterion has no impact on the SEAL calculation."* A criterion whose every recorded answer is SEAL-4 can never bind the minimum — inert for the gate while still contributing to the score. Stated as a property of the derivation, because otherwise an assessor cannot see why such criteria never gate.

**The prevalence statement** (p.10), now quoted rather than asserted:

> It is important to emphasize the prevalence of the SEAL criterion over the Sovereignty Score. In the context of cloud services procurement, the Contracting Authority decides what is the minimum SEAL required. Sovereignty score is used to compare the offers that have reached the minimum required SEAL.

This confirms the "minimum SEAL comes from the tender specification" callout the command already carried — it now rests on the source rather than on my reasoning.

**The p.10 per-objective requirements table**, reproduced with the note that it starts at **SEAL-2 because SEAL-2 was the minimum set for the Commission's own competition**. Without that note, a reader assessing against a lower-floor tender finds no rows and may read the absence as "not assessable" rather than "below the published floor".

## The Commission's own caveat on SEAL-4

From "Lessons learnt" (p.13), attributed and dated (published 1 June 2026), and framed as a limitation they propose to revisit rather than a permanent property:

> The level SEAL-4, however relevant, since it describes the highest level of Sovereignty, is not today relevant in the context of EU Sovereignty considering existing dependence to specific supply chains (chips, hardware). Relaxing the level SEAL-4, at least temporarily, would allow to make more difference between providers, especially when it comes to sensitiveness to hostile take-overs.

Worth carrying: a reader who takes SEAL-4 as a realistic procurement target is working against the framework author's own assessment.

## Terminal output

Overall SEAL now prints **above** the Sovereignty Score, with the score labelled as the award criterion and the SEAL named as the rejection gate — matching the prevalence rule rather than leaving the reader to infer precedence from ordering.

## Verification

```
3 mjs guards                       all pass
```

Every verbatim quote above was checked against the source PDF before being written in. `check-doc-control-resolution.py` stays green — this edits a command that resolves the `DOC-CONTROL-HEADER` marker and does not disturb those lines. `docs/llms.txt` updated (now guarded by `check-llms-txt.py`).

No changes to SOV weight values or SEAL level names — a companion PR owns those.

## Merge order

Companion to the weights and scoring PRs; all three touch the command, template, checklist and CHANGELOG. Suggested order: weights → this → scoring. Whichever lands later needs a rebase, mostly on `## [Unreleased]`.


tractorjuice:
## Code Review

**Verdict:** The substantive rule is right and every quotation is verbatim. One factual error to fix before merge: cell `F2` is cited for the wrong thing, in both the command and the quality checklist. Plus a merge hazard shared with the other two PRs.

### Verified against the primary source

I fetched the Implementation guidance PDF and checked each quotation character by character.

- p.9: "The overall SEAL level is **the lowest SEAL level achieved in any of the objectives**." Confirmed verbatim.
- p.9: "When all responses grant SEAL-4, it must be understood that the criterion has no impact on the SEAL calculation." Confirmed verbatim.
- p.10: the full prevalence paragraph. Confirmed verbatim.
- p.10: the table really does begin at SEAL-2, and the source says "from SEAL-2, which was the minimum level set for the competition". Your note about a reader with a lower tender floor is a genuine improvement, not an embellishment.
- p.13: the SEAL-4 caveat. Confirmed verbatim, and framing it as a limitation the Commission proposes to revisit matches the section's own wording.

The minimum-not-average rule is correct and the checklist's explicit rejection of averaging and modal language is a good way to make it enforceable.

### IMPORTANT: `F2` is the overall SEAL, not the per-objective SEAL

The command says:

> **Per-objective SEAL** = the minimum SEAL across every criterion answered within that objective. The Annex calculator XLSX implements this at cell F2: `="SEAL-"&MIN(H5:H251)`

`H5:H251` spans every answer row in the workbook, all eight objectives, so `F2` is the overall SEAL. I scanned the whole sheet:

```
any formula containing MIN(
   F2 ="SEAL-"&MIN(H5:H251)

F/G/H at the eight objective header rows (4, 45, 76, 102, 133, 169, 195, 231):
   all None
```

There is exactly one `MIN()` in the workbook and no per-objective SEAL cell anywhere. The calculator does not compute a per-objective SEAL at all. Your reasoning is sound, the minimum-of-minimums equivalence holds, and `MIN(H5:H251)` is genuinely a minimum over answered criteria since `H` is `=IF(E<n>,1*F<n>,"")`. Only the citation is attached to the wrong bullet: `F2` belongs on the **Overall SEAL** line, and the per-objective line is a property of the framework that the calculator leaves implicit.

Two places to fix:

- `plugins/arckit-eu/commands/eu-cloud-sovereignty.md`, the "Per-objective SEAL" bullet in Section 3
- `references/quality-checklist.md`, the "Overall SEAL is present" item, which repeats "(Annex calculator cell F2)" for the per-objective claim

Worth fixing rather than letting slide, for two reasons. The command instructs the assessor to "cite them separately", so a wrong cell reference is reproduced into every generated artefact. And #797 cites `F2` correctly for the Overall SEAL, so once both land the same file cites the same cell for two different things.

### Cross-PR: rebase rather than resolve at merge time

All three PRs branch from `f1784387`. I merged them locally in your suggested order. #795 then this PR conflicts in six files, not just the CHANGELOG:

```
.arckit/templates/eu-cloud-sovereignty-template.md
CHANGELOG.md
plugins/arckit-claude/plugins/eu/commands/eu-cloud-sovereignty.md
plugins/arckit-claude/plugins/eu/templates/eu-cloud-sovereignty-template.md
plugins/arckit-eu/commands/eu-cloud-sovereignty.md
plugins/arckit-eu/templates/eu-cloud-sovereignty-template.md
```

The conflicts sit exactly on the lines #795 corrects, so this branch's side of the conflict is the pre-fix text:

```
<<<<<<< HEAD
  - **SEAL-3 Technological Sovereignty**: EU jurisdictions apply, ...
=======
  - **SEAL-3 Digital Resilience**: EU law applicable and enforceable, ...
  - Reproduce the per-objective SEAL-2/3/4 requirements table from Guidance p.10 ...
>>>>>>> pr-796
```

and in the template:

```
<<<<<<< HEAD
### 4.1 SOV-1 Strategic Sovereignty (Weight: 20%)
=======
> **SEAL evidenced**, per objective below, is the minimum SEAL ...

### 4.1 SOV-1 Strategic Sovereignty (Weight: 15%)
>>>>>>> pr-796
```

Taking either side wholesale loses something. Wrong weights are caught by #795's new guard; a wrong or duplicated SEAL-3 name is caught by nothing, and I confirmed markdownlint stays clean with both names present. A rebase on `main` after #795 lands is safer than resolving these by hand.

### MINOR: deduplicate during the rebase

After all three merge, Section 3 of the command carries two Overall SEAL bullets and two "state the formula" bullets, the External References table gains duplicate rows for the guidance and the calculator, and the quality checklist carries two overlapping Overall SEAL items. Yours says "the lowest SEAL evidenced across all eight objectives"; #797's says "the minimum SEAL across all answered criteria". Both are true and equivalent, but #797 adds "not derived from the per-objective SEAL evidenced values alone", which reads as a contradiction of your bullet sitting ten lines above it. Worth collapsing into one derivation stated once.

### POSITIVE

- Quoting the prevalence statement rather than asserting it, and saying so explicitly, is the right instinct given the command already carried that claim on your own reasoning.
- The inert-criterion property is easy to miss and genuinely changes how an assessor reads a SEAL-4 row. Stating it as a property of the derivation rather than a footnote is the right placement.
- Putting Overall SEAL above the Sovereignty Score in the terminal output, so precedence is not left to be inferred from ordering, is a good call and matches p.10.

---

Review verified against: Implementation guidance PDF (13pp) and Annex calculator XLSX, fetched from the Commission URLs in the PR body; workbook inspected with openpyxl.

tractorjuice:
Rebased onto `main` (only `CHANGELOG.md` conflicted — v6.10.0 had consumed the `[Unreleased]` section, so the entry moved to a fresh one), and addressed the review.

**IMPORTANT — `F2` cited for the wrong thing.** Confirmed, and the reasoning holds without needing the workbook: `F2` is `="SEAL-"&MIN(H5:H251)`, and `H5:H251` spans every answer row across all eight objectives — their header rows are 4/45/76/102/133/169/195/231, the same rows this PR series reads the weights from (`D4`/`D45`/…). So `F2` is the **Overall** SEAL. A per-objective minimum would have been `MIN(H5:H44)` and so on, and as you found there is no such cell.

`F2` now sits on the Overall SEAL line, with the row-range reasoning stated so it is checkable. The per-objective line is stated as a property of the framework that the calculator leaves implicit, with an explicit instruction not to attribute it to a cell — the command tells the assessor to "cite them separately", so the wrong reference would have been reproduced into every generated artefact.

Corrected in **four** places, one more than the review found:

- `plugins/arckit-eu/commands/eu-cloud-sovereignty.md` — the Section 3 derivation bullets
- the same file's quality-checks bullet, which repeated the claim
- `plugins/arckit-eu/templates/eu-cloud-sovereignty-template.md` — Section 3.3
- `references/quality-checklist.md` — canonical copy in `arckit-claude`, propagated by `sync-shared-assets.py` to all 31

A fifth, caught by `test_plugin_and_cli_templates_have_identical_content` rather than by me: the fix initially landed on the plugin template only, leaving `.arckit/templates/` — which commands resolve **first** — still carrying the wrong citation. Fixed in its own commit.

**On the cross-PR merge hazard** — this branch now also carries #795's normative corrections (the three permuted weights, and `SEAL-3` renamed to "Technological Sovereignty"). It predates #795, so its side of every shared line was the pre-fix text; merged in the wrong order, or with a conflict resolved the obvious way, that silently reverted them, and markdownlint is clean with either SEAL-3 name. Both sides now make the identical change, so git resolves those lines with no conflict at all.

Green: full CI guard set, 1541 pytest, markdownlint over 3800 files. No conflicts with `main`.

tractorjuice:
Merged as a313b42c, after #795.

Worth recording how the integration went, since it was the point of carrying #795's corrections onto this branch: **not one normative line conflicted.** Every conflict was two independent additions at the same anchor, resolved by keeping both. The SOV weights and the SEAL-3 name were byte-identical on both sides, so git had nothing to ask about — which is exactly the outcome the review's cross-PR warning was aiming at.

One ordering decision worth naming: #795's SEAL-2 grammar note now sits directly after the SEAL definitions it annotates, with this PR's Section 3.0b instruction following it.

Verified before merge on the integrated tree, not just on the branch: all 17 CI guards, `test-csf-weights.mjs`, 1541 pytest, markdownlint over 3800 files.

PR #797, opened 2026-08-14
fix(eu): define the Sovereignty Score basis and ship the framework calculator (#782)

### LINKED ISSUE #782: eu-cloud-sovereignty: Sovereignty Score formula demands Score(SOVn) / Max.Score(SOVn) but never defines them
`/arckit:eu-cloud-sovereignty` states the Sovereignty Score formula and demands its inputs, but never defines how `Score(SOVn)` and `Max.Score(SOVn)` are derived. The model is left to invent a numeric scale, and the resulting percentage is presented as a tender award criterion.

Found reviewing #740 and deliberately left open at merge, because closing it needs a domain answer rather than a code change.

## Where it is

The formula is stated in two places:

- `plugins/arckit-eu/commands/eu-cloud-sovereignty.md:130`
- `plugins/arckit-eu/templates/eu-cloud-sovereignty-template.md:86`

```
Sovereignty Score = Σ over the eight objectives of
                    (Score(SOVn) / Max.Score(SOVn)) × Weight(SOVn)
```

The inputs are then demanded as table columns:

- `plugins/arckit-eu/commands/eu-cloud-sovereignty.md:132` asks for "objective, Score, Max Score, Weight, Weighted Contribution"
- `plugins/arckit-eu/templates/eu-cloud-sovereignty-template.md:104` ships that table

## Why it matters

What the command actually collects per objective is a **SEAL level** (claimed versus evidenced) plus checkbox evidence rows against the contributing factors. Nothing anywhere maps a SEAL level, or the evidence, onto `Score(SOVn)`. Nothing states what `Max.Score(SOVn)` is.

So the model has to invent a scale. The most likely guess is `Score = SEAL evidenced, Max.Score = 4`, but that is a guess, and a different plausible guess (counting satisfied contributing factors, say) yields a materially different percentage for the same evidence.

That number is not decorative. It surfaces as:

- the Executive Summary figure,
- `Sovereignty Score: {N}% (award-criterion contribution)` in the terminal output (`commands/eu-cloud-sovereignty.md:194`),
- a success criterion the command checks itself against (`:249`).

The command is explicit that this contributes to the tender's quality score as an award criterion (`:133`). An invented scale therefore produces a fabricated, non-reproducible number that carries the framework's authority into a live procurement evaluation, and two runs over identical evidence can disagree.

This is the same failure class the rest of the command is careful to prevent. It already refuses to invent minimum SEAL levels, refuses to name providers, and flags self-declared SEALs as unverified. The scoring scale is the one place where invention is currently not just permitted but required.

## What needs deciding

This is a question about the framework, not about ArcKit, which is why it is an issue rather than a patch:

1. **Does the EU Cloud Sovereignty Framework v1.2.1 define a per-objective scoring scale?** If it does, state it explicitly in the command and the template, alongside the weights and the SEAL definitions, and have the command reproduce it verbatim the way it already reproduces the five SEAL definitions in full.
2. **If it does not**, the command should say so and stop computing a percentage. Emitting per-objective SEAL evidenced against minimum SEAL, with no aggregate, is honest and still useful. A score presented without a defined scale is worse than no score.

Either resolution should also drop or reword the `Sovereignty Score` line in the terminal summary block accordingly.

## Note

`plugins/arckit-claude/references/quality-checklist.md` already carries a checklist item flagging this, added when #740 merged:

> Sovereignty Score computed from the stated formula, and the per-objective Score / Max Score basis is stated rather than assumed

That surfaces the gap to a reviewer but does not close it.

/cc @umag, since this needs your read of the framework text.

umag:
Closes #782.

You are right that the formula's inputs are undefined and that the model must invent a scale. Neither resolution the issue offers turns out to be the answer: the framework **does** define the scale, and the guess the issue flags as most likely — `Score = SEAL evidenced, Max.Score = 4` — is definitively wrong.

Source: the [Annex — Sovereignty assessment calculator](https://commission.europa.eu/document/download/3acb8fe8-8a4a-4339-ae74-f56138d913d1_en?filename=Annex+-+Sovereignty+assessment+calculator.xlsx) accompanying the [Implementation guidance](https://commission.europa.eu/document/download/2ad80a48-166f-4c77-a513-80c53ca2a128_en?filename=Cloud+Sovereignty+Framework+-+Implementation+guidance.pdf).

## What the inputs actually are

- **`Score(SOVn)`** — the sum of the *selected answers' values* for that objective's criteria (calculator `E4=SUM(E5:E44)`, and so on per objective).
- **`Max.Score(SOVn)`** — a **nominal 1000 for every objective**, by construction: each objective normalises its criteria so they total 1000. SOV-1 8×125, SOV-2 6×167, SOV-3 5×200, SOV-4 6×167, SOV-5 7×143, SOV-6 5×200, SOV-7 7×143, SOV-8 4×250 — **48 criteria** in total.

Once `Max.Score` is known to be uniform, the formula we already state is algebraically identical to the calculator's own top-level cell:

```
E1 = ((D4*E4 + D45*E45 + D76*E76 + D102*E102 + D169*E169 + D133*E133 + D195*E195 + D231*E231)) / 1000
```

So the formula was never wrong — one definition was missing, and it was the one that made the rest unusable.

## SEAL is not an input to the score

Guidance p.9:

> The same answers are used to determine the SEAL of each row, with each answer defining the SEAL level of the question.

Score and SEAL are two **parallel readings of the same 48 answers**; neither derives from the other. That is why the command's existing inputs — SEAL claimed vs SEAL evidenced, plus checkbox evidence — cannot produce a score at all, and why `Score = SEAL, Max = 4` would have been a plausible, reproducible-looking, wrong number. The command now states this explicitly, since it is the misreading the formula invites.

## 1000 is design intent, not arithmetic

The workbook rounds answer values to 2dp, so the actual per-objective maxima are:

```
1000.03 · 1002.00 · 1000.00 · 1002.00 · 1001.00 · 1000.00 · 1001.00 · 1000.00
```

Under the framework's own `/1000` divisor a **maximal response scores 100.0756%**, not 100%. That is pinned in the test rather than clamped away — reporting the framework's behaviour faithfully matters more than a tidy ceiling, and a reader who sees 100.08% should be able to find out why.

(The workbook also rounds each *selected* answer to an integer before summing — `D9=125.01` against `E9=125`, `D12=62.5` against `E12=63` — which is why its partial SOV-1 example reads 533 where precise values give 530.93. Recorded as an informational discrepancy, not asserted equal. Under per-answer rounding the maximal score is 100.075%, so the overshoot holds either way.)

## What ships

The **48-criterion catalogue** as versioned data at `plugins/arckit-eu/data/csf-criteria-calculator-2026-06-01.json`, with provenance in-file, plus `plugins/arckit-claude/scripts/csf-score.mjs` so the score is computed rather than invented. The command still requires the per-criterion arithmetic to appear in the artefact, so the result is checkable by hand where the scorer is not distributed — Codex, Gemini, OpenCode and Copilot targets receive the catalogue but not the script.

**Reuse verified before vendoring.** Commission Decision 2011/833/EU applies; Commission documents are reusable under **CC BY 4.0** with credit given and changes indicated. Attribution, licence basis and the change-indication (workbook transcribed to JSON) are carried in `provenance.reuse`.

**Named for the artefact, not the framework version.** The guidance narrative describes **43** questions for the first tender wave; the published calculator scores **48**. Both figures are recorded in the file with the discrepancy stated rather than silently reconciled — which is also why the filename carries the calculator's publication date instead of `v1.2.1`.

The framework's own caveats are recorded too: answer values *"can be adapted by contracting authorities"*, the weights column is headed *"Score (examples)"*, and the workbook's example column holds *"fictitious values … for the sole purpose of exemplification"* — never presented as a real provider assessment.

## Distribution

`plugins/arckit-eu/data/` is a directory shape no overlay had, so three registries needed resolving:

- **`pyproject.toml`** — two `shared-data` entries, mirroring the `generate-document-id.mjs` / `config` precedent
- **`scripts/converter.py`** — a generic `data/` merge in `copy_extension_files()`, parallel to the existing `templates/` merge, so any future plugin's `data/` is covered rather than special-casing this one
- **`scripts/sync-claude-plugin-layout.py`** — no change needed; `expected_files()` already walks the tree via `rglob`

**And a fourth that nothing would have reported.** A blanket `data/` rule in `.gitignore` — written for the generated `extensions/*/data/` converter output — silently swallowed the new tracked source directories. The PR would have shipped without the catalogue while every local check passed. Scoped negations added; generated `extensions/*/data/` stays ignored.

## Verification

```
test-csf-score.mjs                 PASS
4 mjs guards                       all pass
```

`test-csf-score.mjs` asserts what is genuinely checkable rather than claiming an oracle the workbook cannot provide: weights sum to 1.0; per-objective maxima exact; criterion counts 8/6/5/6/7/5/7/4; all 235 answers carry a SEAL in 0–4; the maximal response is 100.0756% and not clamped; the overall SEAL is a minimum and not an average. Wired into `lint-markdown.yml` with its own `run:` step.

One incidental fix: the command body said "Claude Code" literally, which propagated into the generated Codex skill and failed `test_codex_skills_do_not_expose_claude_command_syntax_or_hooks`. Reworded to the repo's existing "non-Claude runtimes" convention.

## Scope

No SOV weight values or SEAL level names changed in prose — companion PRs own those. Only the new data file carries the corrected weights, since it transcribes the calculator directly.

## Merge order

Third of three. The weights PR should land first; its values are load-bearing for this catalogue.


tractorjuice:
## Code Review

**Verdict:** The transcription is excellent and I verified it answer by answer against the published workbook. One blocker: the catalogue never reaches CLI-scaffolded Codex and OpenCode projects, so #782 stays unfixed for exactly the runtimes this PR says receive it. Two minors, plus the merge hazard shared with #795 and #796.

### Verified against the workbook, not the PR body

I downloaded the Annex calculator and diffed every field of `csf-criteria-calculator-2026-06-01.json` against it with openpyxl.

```
criteria per objective   8 / 6 / 5 / 6 / 7 / 5 / 7 / 4   =  48   match
answers                  235                                     match
per-objective maxima     1000.03 1002.00 1000.00 1002.00
                         1001.00 1000.00 1001.00 1000.00         match
weights                  D4=0.2 D45=0.1 D76=0.1 D102=0.15
                         D133=0.1 D169=0.15 D195=0.15 D231=0.05  match
answer values, SEAL levels, criterion descriptions, objective
descriptions                                                     match
```

Only four deltas, all of them correct handling of the source rather than transcription error:

- SOV-1 criterion 3 and criterion 7 each carry a stray workbook row with no value (one literally labelled `0`). The catalogue omits both. Right call, they are spreadsheet noise and are visible as artefacts in the guidance screenshots on p.8 and p.9.
- SOV-1 criterion 6 rows 31 and 33 have blank label cells in the workbook. The catalogue records them as `label: null`.

The rest of the reasoning holds too. `E1` is reproduced faithfully including its out-of-order terms (`D169` before `D133`). The `Score (examples)` caption is on p.7. `I4` in the workbook reads "column E 'score' has been filled with fictitious values ... for the sole purpose of exemplification", so your caveat is quoted, not paraphrased. p.13 gives the 43-question figure, so the 43-versus-48 discrepancy is real and recording both is right. And the overshoot arithmetic is exact:

```
0.2(1000.03) + 0.1(1002) + 0.1(1000) + 0.15(1002)
+ 0.1(1001) + 0.15(1000) + 0.15(1001) + 0.05(1000)  =  1000.756
/1000  ->  100.0756%
```

I ran the scorer on a synthesised maximal response and got `100.0756%` and `SEAL-4`, matching. `test-csf-score.mjs` passes, and its assertions are the honest ones: it tests invariants the workbook can actually adjudicate rather than claiming an oracle it cannot.

Repo checks all reproduce your numbers: `1512 passed, 225 skipped`, converter zero drift, `sync-shared-assets --check`, `sync-claude-plugin-layout --check`, `check-guide-parity`, `check-llms-txt`, `check-doc-type-registry` and `check-doc-control-resolution` all clean. The data file reaches all seven extensions with paths rewritten per target.

### BLOCKER: `arckit init` does not scaffold `.arckit/data/`

The generated Codex skill and OpenCode command both instruct the model to read:

```
.arckit/data/csf-criteria-calculator-2026-06-01.json
```

and the command calls that file "the ONLY source for `Score(SOVn)`, `Max.Score(SOVn)`, and each criterion's SEAL level", with "Never invent, estimate, or paraphrase these values".

That path is never created. Run against this branch with the installed CLI:

```
$ arckit init /tmp/initest --ai codex --no-git
$ find /tmp/initest/.arckit -maxdepth 1 -mindepth 1 -type d
/tmp/initest/.arckit/templates
/tmp/initest/.arckit/schemas
/tmp/initest/.arckit/scripts
/tmp/initest/.arckit/templates-custom
/tmp/initest/.arckit/references
/tmp/initest/.arckit/config

$ ls /tmp/initest/.arckit/data
ls: cannot access '/tmp/initest/.arckit/data': No such file or directory
```

Same for `--ai opencode`. So on a CLI-scaffolded project the model is pointed at a file that does not exist, and falls back to inventing the scale, which is #782 verbatim. The plugin path is fine, `${CLAUDE_PLUGIN_ROOT}/data/` resolves.

The cause is that `get_data_paths()` uses an explicit per-directory key list, and `data` was not added to it. `codex_references`, `codex_schemas` and `codex_config` each have an entry and a matching copy step; there is no `codex_data`. Suggested fix, mirroring `codex_schemas` exactly:

- `src/arckit_cli/__init__.py:215` add `"codex_data": base_path / "extensions" / "arckit-codex" / "data"`
- alongside the copy at lines 756 to 761, copy it to `project_path / ".arckit" / "data"`
- add `"data"` to `REQUIRED_ASSETS["extensions/arckit-codex"]` in `hatch_build.py`, so a hollow build cannot ship the extension without it

The PR body works through three registries plus the `.gitignore` rule it calls "a fourth that nothing would have reported". This is a fifth, and it has the same signature: `data/` reaches the extension trees but not the scaffolded project tree, and every local check passes regardless.

### MINOR: the two blank labels are indistinguishable from a transcription loss

`label: null` on the two SOV-1 criterion 6 answers worth 31.25 and 93.75 is faithful, the workbook really does have blank cells there. But an assessor reading the catalogue, or Appendix A generated from it, sees two unlabelled options and has no way to tell whether that is the Commission's gap or yours. A short note in the criterion entry, or in `$comment`, would settle it. The existing `questionCount.note` is a good model for how to phrase it.

### MINOR: the publication date is load-bearing but uncorroborated

`provenance.publicationDate: "2026-06-01"` names the file and is quoted as "published 1 June 2026" in #796's attribution. I could not corroborate it from either artefact: the workbook's `docProps` gives created `2026-05-28T09:59:55Z`, and the PDF's `CreationDate` is `2026-06-03`. The date sits plausibly between them and presumably comes from the Commission page, but since you deliberately named the file by publication date rather than framework version, it is worth recording where the date comes from in `provenance`.

### Cross-PR: rebase rather than resolve at merge time

Merging all three locally in your suggested order, this PR conflicts with the merged state of #795 and #796 in eight files, including two the PR body does not anticipate:

```
.github/workflows/lint-markdown.yml     (both PRs add a run: step at the same point)
docs/llms.txt                           (both rewrite the same EUCSF line)
```

More importantly, because this branch also predates #795, its side of the command and template conflicts carries the pre-fix SEAL wording and the pre-fix `(Weight: 15%)` headers. So the corrected normative text is presented as a discardable merge side a second time. The weight regression is caught by #795's guard, which fired when I introduced one by accident. A duplicated or reverted SEAL-3 name is caught by nothing, and markdownlint stays clean with both names present.

A rebase on `main` after #795 and #796 land is safer, and gives an opportunity to collapse the duplicated Overall SEAL and "state the formula" bullets and the duplicate External Reference rows that otherwise end up in Section 3.

### POSITIVE

- Establishing the reuse basis before vendoring Commission material, and carrying the attribution and change-indication in `provenance.reuse`, is exactly right and worth copying for any future vendored source.
- Naming the file by the calculator's publication date rather than the framework version, because the two disagree on the question count, is a good judgement call.
- Refusing to clamp 100.0756% to 100 and pinning it in the test is the right choice. So is recording the 533 versus 530.93 gap as informational rather than asserting equality.
- Making the `data/` merge in `converter.py` generic rather than special-casing `arckit-eu` means the next plugin gets it free.

---

Review verified against: Annex calculator XLSX inspected with openpyxl and diffed field by field against the shipped JSON; Implementation guidance PDF (13pp); `arckit init` run for both `--ai codex` and `--ai opencode`; full test suite, converter and all sync guards run on this branch.

tractorjuice:
Rebased onto `main` (only `CHANGELOG.md` conflicted — v6.10.0 had consumed the `[Unreleased]` section), and addressed the review.

**BLOCKER — `arckit init` does not scaffold `.arckit/data/`.** Confirmed and fixed. `get_data_paths()` builds an explicit per-directory key list and `data` had no entry, so the model was pointed at a file that was never created and fell back to inventing the scale — #782 verbatim, on exactly the runtimes this PR says receive the fix.

One correction to the suggested placement: copying it alongside the `codex_schemas` block would not have fixed it, because that block sits inside `if ai_assistant == "codex" or all_ai:` — and `--ai opencode` was broken too, as you found. The converter rewrites the path to `.arckit/data/` for **all three** CLI targets, so it is copied unconditionally next to `references`, which is already handled that way for the same reason. Verified end to end:

```
codex      .arckit/data -> csf-criteria-calculator-2026-06-01.json (60059 bytes)
opencode   .arckit/data -> csf-criteria-calculator-2026-06-01.json (60059 bytes)
copilot    .arckit/data -> csf-criteria-calculator-2026-06-01.json (60059 bytes)
```

`"data"` added to `REQUIRED_ASSETS` in `hatch_build.py` as suggested, verified by removing the directory:

```
missing_assets(): []
missing_assets() without data/: ['extensions/arckit-codex/data']
```

No `pyproject.toml` change needed — `extensions/arckit-codex` ships wholesale to `share/arckit/`, so `data/` is included once it exists. The guide now also records where the catalogue lives per install method.

**MINOR — blank labels indistinguishable from a transcription loss.** Agreed. SOV-1 criterion 6 now carries a `note` saying rows 31 and 33 are the workbook's own blank cells, transcribed as-published rather than back-filled, modelled on `questionCount.note` as you suggested.

**MINOR — publication date uncorroborated.** Agreed and recorded as `provenance.publicationDateSource`: the Commission's download page, with both conflicting metadata values written down (workbook `docProps` created `2026-05-28T09:59:55Z`, guidance PDF `CreationDate` `2026-06-03`) and why the date is load-bearing — it names the file, chosen over the framework version because the two disagree on the question count.

**Cross-PR** — this branch now carries #795's corrections too, so both sides of every shared line make the identical change. The one remaining conflict hunk that still contains SOV weight rows has the **correct** weights on both sides; the only delta there is this PR's own `[Max]` → `1000`. So no merge order can revert them.

Green: full CI guard set, 1541 pytest, `test-csf-score.mjs`, converter zero drift, markdownlint over 3800 files. No conflicts with `main`.

tractorjuice:
Merged as 0666b790, last of the three. Closes #782.

Integrating this one needed four real decisions rather than a straight union, worth recording:

1. **Both sides stated the Sovereignty Score formula**, and `main`'s was the unresolved `/ Max.Score(SOVn)` form this PR exists to replace. Kept the resolved one and dropped the other — a union would have shipped the defect alongside its own fix, in the same section.
2. **External References** listed the same two sources on both sides with different descriptions. Merged the descriptions into one row each rather than duplicating the guidance and the calculator.
3. **The Scored Result table** appeared on both sides with identical, correct weights; kept this PR's nominal `1000` Max column.
4. **`docs/llms.txt`** — #796 and this PR each rewrote the same entry, so neither side alone was complete. Combined, so it now carries both the Overall SEAL rule and the Score/Max.Score basis.

The normative lines did not conflict at all, because this branch already carried #795's corrections.

Verified on the integrated tree after merge: 22 CI guards, `test-csf-weights.mjs`, `test-csf-score.mjs`, 1541 pytest, markdownlint over 3800 files, converter zero drift. And the blocker re-checked end to end against merged `main` — `arckit init --ai opencode` now lands `csf-criteria-calculator-2026-06-01.json` in `.arckit/data/`.

Issue #466, opened 2026-05-13
Follow-up to #442: remaining financial-services adoption items


tractorjuice:
Tracker for the items from #442 (closed 2026-05-13 as completed) that have **not yet shipped**. Splitting these out per close-and-resplit so each has a clean acceptance check rather than being buried inside a meta-tracker.

## Shipped already (under #442)

- ✅ Item 1 partial — three-tier reader/orchestrator/writer split for `datascout`, `grants`, `gov-reuse` (PR #446)
- ✅ Item 2 — schema-gated handoffs (3 schemas + `validate-handoff.mjs` + 21 fixtures + 3 Node test runners in CI)
- ✅ Item 3 — reader output schemas (shipped alongside the 3 split commands)
- ✅ Item 4 — cross-reference linter (`scripts/check_references.py`, PR #464)
- ✅ Item 15 — MCP catalogue (`docs/MCP-CATALOGUE.md`, PR #465)
- ✅ Item 18 — tool allowlist by default on all 10 research agents (PR #445) — **silently reverted on 3 agents by PR #446; restored and CI-guarded in PR #807** ([detail](https://github.com/tractorjuice/arc-kit/issues/466#issuecomment-5340538175))

### High impact

- [x] **Item 1 remainder** — extend reader/writer split to the other research agents: **COMPLETE**
  - [x] `arckit-research` (vendor packs, web pages — biggest injection surface) — **done, PR #808**
  - [x] `arckit-aws-research` (AWS MCP responses) — **done, PR #812**
  - [x] `arckit-azure-research` (Microsoft Learn MCP responses) — **done, PR #812**
  - [x] `arckit-gcp-research` (Google Developer Knowledge MCP responses) — **done, PR #812**
  - [x] `arckit-gov-code-search` (govreposcrape responses) — **done, PR #810**
  - [x] `arckit-gov-landscape` (govreposcrape responses) — **done, PR #811**
  - [x] (`arckit-framework` — confirmed skip: synthesis-only, no external input)
- [ ] **Item 5** — Managed Agents API headless deployment track (strategic; open question — needed once governance-on-cron is a real ask)

### Medium

- [x] **Item 9 + Item 21** — README regulatory disclaimer + concrete "does NOT do" enumeration — **done, PR #809**
[...]

Issue #447, opened 2026-05-06
Datascout three-tier flow: regression in non-Claude runtimes (Codex/Gemini/OpenCode/Copilot/Paperclip)


tractorjuice:
## Background

PR #446 (v4.16.0–v4.17.1) split `arckit-datascout` into a three-tier subagent architecture (orchestrator slash command → reader subagent + writer subagent) with JSON-Schema-validated handoff between tiers. The pattern is faithful to `anthropics/financial-services` and works correctly in the Claude Code plugin (manually tested in test repo v48 — passed).

## The regression

Pre-PR, ArcKit's converter inlined a single-agent `arckit-datascout.md` into all six distribution formats. Each non-Claude target had a working datascout that did discovery + scoring + writing inline.

Post-PR, the source agent file was deleted (commit `84f11fac`) and orchestration logic moved to the slash command body (`arckit-claude/commands/datascout.md`). The slash command body now contains heavy "dispatch the reader subagent via the `Agent` tool" instructions that the converter inlines into all non-Claude targets:

| Runtime | Files generated | What the LLM sees | Works? |
|---|---|---|---|
| Claude Code (plugin) | `commands/datascout.md` (orchestrator) + `agents/arckit-datascout-{reader,writer}.md` | Slash-command in main thread dispatches `Agent` to spawn reader/writer subagents | ✅ |
| Codex CLI | `arckit-codex/agents/arckit-datascout.md` (303 lines, slash-command body inlined) + `skills/arckit-datascout/SKILL.md` + `prompts/arckit.datascout.md` | Same orchestrator prompt with subagent-dispatch instructions | ❌ |
| Gemini CLI | `arckit-gemini/agents/arckit-datascout.md` + `commands/arckit/datascout.toml` | Same | ❌ |
| OpenCode CLI | `arckit-opencode/agents/arckit-datascout.md` + `commands/arckit.datascout.md` | Same | ❌ |
| GitHub Copilot | `arckit-copilot/agents/arckit-datascout.agent.md` + `prompts/arckit-datascout.prompt.md` | Same | ❌ |
| Paperclip (TS plugin) | `arckit-paperclip/...` JSON entry inlining slash-command body | Same | ❌ |

The reader and writer source files (`agents/arckit-datascout-{reader,writer}.md`) are correctly *filtered* from non-Claude by the `subagent: true` flag — they don't propagate. The leak is the slash-command body, which non-Claude runtimes inline as their datascout entry but cannot honour.

[...]

PR #824, opened 2026-08-21
fix: ship single-agent monoliths for tenders/competitors on non-Claude targets (#447)

tractorjuice:
Closes the executable half of #447.

## The regression

`/arckit:tenders` and `/arckit:competitors` were born three-tier in #558 (v5.9.0, 2026-06-02) and never got the pre-split monolith that `converter.py` swaps in when generating non-Claude targets. With no `agents/arckit-{name}.md`, `build_agent_map()` has nothing to substitute and the converter falls through to the orchestrator body. All seven extensions have therefore been shipping:

> Dispatch the reader using the `Agent` tool with `subagent_type: "arckit-tenders-reader"`

which Codex, Gemini, OpenCode, Copilot, Paperclip, Vibe and Kimi cannot honour. Verified across every target before the fix; two leaking skills, zero after.

This is the same defect #447 reported against `datascout` in #446, fixed there by retaining that command's monolith. It recurred because nothing derived the rule: `tests/codex/test_codex_extension.py` checked three hardcoded skill names, on one target, so two commands added afterwards were never in scope.

**Claude Code is unaffected** and keeps the three-tier path.

## Changes

**1. Two single-agent monoliths.** `agents/arckit-tenders.md` and `agents/arckit-competitors.md` do the MCP querying, derivation and rendering in one role, carrying the same guardrails, 15-call MCP budget, `query_sql`-never-granted rule, mandatory awarded-value caveat, `DOC-CONTROL-HEADER` resolution and quality-checklist step as the split tiers. The competitors monolith keeps the bounded vendor-profile enrichment.

**2. `scripts/check-orchestrator-leaks.py`**, wired into `lint-markdown.yml`. Enumerates every `plugins/*/commands/*.md` and resolves each through `converter.py`'s own `build_agent_map()`, so it cannot drift from what the converter emits and covers a new split command the day it lands. Also checks `commands-standalone/` overrides. Its regexes tolerate backticks — the old test's plain-substring `"Agent tool"` never matched the source's ``the `Agent` tool``, so it was under-firing on top of being under-scoped.

Verified to fail correctly:

```
$ mv plugins/arckit-claude/agents/arckit-{tenders,competitors}.md /tmp/ && python3 scripts/check-orchestrator-leaks.py
Orchestrator-leak check FAILED:
  - plugins/arckit-claude/commands/competitors.md: orchestrator prose (...) with no pre-split monolith ...
  - plugins/arckit-claude/commands/tenders.md: orchestrator prose (...) with no pre-split monolith ...
2 problem(s) found across 186 generated prompt(s).
```

**3. Vibe subagent filter.** `generate_vibe_agent_toml_files()` was the one generation path missing `is_subagent_file()` — every other target had it. Vibe shipped 29 agent definitions, 19 of them reader/writer subagents whose own descriptions read "Not user-invocable — only the orchestrator dispatches this subagent via the Agent tool", while Vibe surfaces every agent file as a user-invocable `--agent` target and has no orchestrator-dispatch primitive. Now 12, matching the other targets.

**4. De-hardcoding.** `CLAUDE_ONLY_COMMANDS` promoted from a function-local literal to a module constant in `converter.py` (one copy, shared with the guard). `AGENT_BACKED_SKILLS` derived from the plugin source. The Codex leak test enumerates all skills. The two Vibe tests that asserted the leak was correct behaviour are rewritten to assert the opposite, with the reasoning recorded.

**5. Docs.** Guide note on both commands' Agent Architecture blockquotes naming the non-Claude fallback (root canonical, synced). `CLAUDE.md`: 29 → 31 agents, two table rows, monolith list extended, and the #447 lesson recorded alongside the #466 one.

## Verification

- `pytest tests/` — 1546 passed, 225 skipped
- All 10 relevant `check-*.py` guards pass; `sync-shared-assets.py --check` clean
- `markdownlint-cli2` clean on all changed markdown
- Post-converter sweep across all seven generated trees: zero orchestrator prose in any command, prompt, skill or agent file

## Not in this PR

Two further #447 findings, both raised on the issue and better handled separately:

- **MCP manifest coverage is uneven.** Against the plugin's six servers: Codex 6/6, Kimi 6/6, Gemini 5/6, Vibe 5/6, OpenCode 3/6. These are hand-maintained manifests, not converter output. A tenders monolith on Gemini, Vibe or OpenCode has no `uk-tenders` server to call, and `datascout`'s existing fallback on OpenCode is already missing two of its three.
- **Phase 2 (three-tier on Codex)** needs its runtime research redone — the issue costed it against Codex v0.128.0 and Gemini CLI v0.41.1, both months stale.

https://claude.ai/code/session_01CMhNeNM12vHqHzVXNNhwCH


tractorjuice:
## Code Review

**Verdict:** Approve — the fix verified end-to-end on the branch; two nits, nothing blocking.

### Verified (ran locally, not just CI)

- `scripts/check-orchestrator-leaks.py` passes: **186 generated prompts clean**.
- `python scripts/converter.py` regenerates with **zero tracked-file drift**.
- `tests/codex/` + `tests/vibe/` — **74 passed** against the freshly regenerated extensions.
- `check-doc-type-registry.py`, `sync-shared-assets.py --check`, `sync-claude-plugin-layout.py --check`, `check-guide-parity.py` all pass on the branch.
- Monolith guardrail claims spot-checked in both agents: 15-call MCP budget (`arckit-tenders.md:127`, `arckit-competitors.md:120`), mandatory awarded-value caveat, `DOC-CONTROL-HEADER` resolution step, Write-tool discipline, and `query_sql` appears only as a prohibition (`arckit-tenders.md:253,262`). Frontmatter fields all valid; MCP allowlists use the `mcp__plugin_arckit_uk-tenders__*` prefix correctly.

### MINOR

1. `scripts/check-orchestrator-leaks.py` — the `p != MIRROR_DIR` filter in `plugin_dirs()` is dead code: `PLUGINS_DIR.iterdir()` yields only direct children of `plugins/`, and `MIRROR_DIR` (`plugins/arckit-claude/plugins`) is a grandchild, so the comparison can never exclude anything. Drop it (keeping the comment), or anchor the exclusion somewhere it could actually fire.
2. Merge-order note with #825: the two PRs touch different regions of `converter.py` so git merges clean, and I've verified the incoming `plugins/arckit-oaa/commands/*.md` contain none of this guard's forbidden vocabulary — whichever lands second just needs the normal CI re-run.

### POSITIVE

- The guard deriving its scope through `converter.py::build_agent_map()` instead of restating the rule is exactly the right fix for *how* #447 recurred; same for `AGENT_BACKED_SKILLS` now being derived from the plugin source in the codex test.
- `CLAUDE_ONLY_COMMANDS` promoted to a single module constant shared by the converter and the guard — one copy, as the repo convention demands.
- The Vibe fix removes 19 dead "Not user-invocable" agents from a target that surfaced every agent file as a user-invocable `--agent` option.

---

https://claude.ai/code/session_01X6mjxfQb9qrWrxaBRkDpQw

PR #850, opened 2026-09-03
Add component provenance and value-chain visibility validation

tractorjuice:
## Summary

Adds two new validation rules to Wardley Map documents that enforce governance over component sourcing and visibility alignment with the project's value chain. These rules prevent fabricated document references and ensure consistency between a map and its source artefacts.

### New Module: `wardley-provenance.mjs`
A pure, side-effect-free rules engine that implements:

1. **Component Provenance** — validates that every component in a Component Inventory table with a `Source` column cites something that actually exists:
   - Citation IDs declared in the map's own Citations table
   - Document IDs or doc-type codes the project holds (e.g., `ARC-001-WVCH-v1.0`, `WVCH`)
   - External Doc IDs listed in the map's Document Register
   - The literal `Assumption` for analytical judgement with no source document

2. **Value-Chain Visibility Join** — when a component is sourced to the project's WVCH artefact, its visibility must match the value chain's own number (within 0.005 tolerance). The value chain owns visibility; the map owns evolution.

Both rules are **claim-scoped**: tables without a `Source` column are not checked (backward compatible with pre-column maps), and visibility is joined only for rows that explicitly cite the value chain, allowing maps deliberately anchored on different user needs to position components differently.

Exports parsing functions (`parseInventoryRows`, `parseValueChainInventory`, `parseCitations`, `parseRegisterDocIds`), resolution logic (`resolveSource`), and the main check function (`checkProvenance`).

### Updated: `validate-wardley-math.mjs`
Integrates the new provenance rules into the existing validation hook:

- Walks the project filesystem to discover artefact document IDs and doc-type codes
- Identifies the highest-versioned WVCH artefact as the value chain
- Reads the value chain's Component Inventory and the map's Citations/Document Register
- Calls `checkProvenance()` and reports errors alongside existing math checks
- Gracefully degrades: filesystem read failures or parse errors produce no findings rather than blocking writes

### Test Suite: `wardley-provenance.test.mjs`
Comprehensive tests covering:

- Parsing of Component Inventory, value chain, citations, and document registers
- Component name normalization across case, quoting, and emphasis
- Source resolution for all valid forms (citations, doc IDs, doc-type codes, assumptions)
- Rejection of empty, placeholder, and fabricated sources
- Provenance and visibility-join rule enforcement
- End-to-end hook integration with temporary project fixtures

### Documentation & Templates
- Updated `wardley-map-template.md` (both `.arckit/` and `plugins/arckit-claude/` copies) to include the `Source` column in all Component Inventory tables with guidance on valid values
- Added "Where Each Component Came From" section to `docs/guides/wardley.md` and `plugins/arckit-claude/docs/guides/wardley.md` explaining the Source column, valid references, and the visibility-join rule
- Updated `plugins/arckit-claude/commands/wardley.md` to document the Source column requirement
- Updated `plugins/arckit-claude/hooks/README.md` to describe the two new validation rules
- Updated `plugins/arckit-claude/docs/ENFORCEMENT.md` and `plugins/arckit-claude/skills/wardley-mapping/SKILL.md` to reference the new checks

## Implementation Details

- **Pure design**: `wardley-provenance.mjs` has no filesystem or process dependencies; `validate-wardley-math.mjs` handles all I/O and feeds results in
- **Markdown table parsing**: Robust handling of Markdown pipes, separators, and header detection; skips template placeholders and Evolution Analysis tables
- **Filesystem walk**: Best-effort discovery of project artefacts up to 3 levels deep; unreadable projects degrade gracefully to no-op checks
- **Version selection**: When multiple WVCH versions exist, the highest (compared numerically, not lexically) is used as the authoritative value chain
- **Error messages**: Actionable hints guide users to fix provenance issues (cite

https://claude.ai/code/session_01AD6w6265bYoZSRWedNqmDi


tractorjuice:
Review of the first cut found ten confirmed issues; all are fixed in the follow-up commit on this branch.

**The one that mattered most.** The command and template primed the Source as `ARC-nnn-WVCH-v1.0`, but WVCH is multi-instance and `/arckit:wardley.value-chain` writes `ARC-001-WVCH-001-v1.0.md`, so the first real run would have blocked every value-chain-sourced row. The e2e tests passed only because the fixture created the unsequenced name. Both are corrected, and every fixture now uses the real filename.

**Correctness of the two rules**
- The visibility join follows the value chain the row cites (a project may hold several); a bare `WVCH` source means the project's own current chain, highest sequence then version.
- The whole `projects/` tree is indexed, so `ARC-000-PRIN-v1.0` and a sibling project's REQ resolve; hyphenated doc types resolve via the registry; `ARC-001-WVCH-001` without a version resolves to the highest version on disk.
- A Citations or Register row naming an ARC document must point at a real file, so a self-authored row can no longer launder an invented ID.
- A value chain that is still the template blocks a row citing it, with a reason; `{0.00}` is not trusted; escaped pipes, code spans and bold headers are handled.

**Safety of the hook**
- No I/O at all unless the map has a Source-bearing row (WVCH/WGAM/WCLM writes are untouched).
- The walk skips `external/`, `vendors/`, `policies/`, uses an accumulator, and sits inside the fail-soft guard with the WVCH reads, so a large tree can neither block nor fail open, and the five math checks are never lost.
- An unreadable `projects/` root skips provenance rather than reporting every real artefact as fabricated.

**Claims match runtimes.** The command, template and guide now say the block is the Claude Code plugin's and a rule the other runtimes hold themselves; the converter output no longer promises a block six runtimes cannot make. ENFORCEMENT.md rows and hooks/README.md describe the actual behaviour.

Tests: 40 in `wardley-provenance.test.mjs` (from 21), covering each finding; full JS suite 396 pass; template consistency, guide parity, references, lint and the Python suite clean.

PR #851, opened 2026-09-03
Add /arckit:archify command for interactive HTML diagram rendering

### LINKED ISSUE #826: add archify skill with command or parameter for diagrams
### Summary

/arckit:archify --parameter

### Problem / use case

https://tt-a1i.github.io/archify/

### Proposed solution

https://tt-a1i.github.io/archify/

### Alternatives considered

_No response_

### AI Tool

Claude Code

### Additional context

_No response_

tractorjuice:
## Summary

Closes #826. Introduces `/arckit:archify`, which renders ArcKit's governed artefacts as interactive, self-contained HTML diagrams — inline SVG, dark/light themes, pan/zoom, click-to-trace dependencies — opening offline with **no external requests**. Two engines sit behind one command.

| Requested type | Engine | Availability |
|---|---|---|
| `architecture`, `workflow`, `sequence`, `dataflow`, `lifecycle` | [Archify](https://github.com/tt-a1i/archify) (third party, MIT) | Only when the user installs it |
| `wardley` | ArcKit's own `owm-to-html.mjs` | Always — ships with the plugin |

## Design decisions

**Archify is bridged, not vendored.** `scripts/archify-detect.mjs` locates a user-installed copy at run time; when absent, the command says so and hands off to `/arckit:diagram` rather than fabricating an artefact. Vendoring was rejected because it is ~7.5MB that `sync-shared-assets.py` and `converter.py` would multiply across ten community plugins and seven extensions, it moves fast enough (v2.17-dev) that a fork would drift within weeks, and its skill contract includes an upstream update check ArcKit should not inherit on a user's behalf. Its `validate`/`deliver` loop was exercised end-to-end on ArcKit-shaped content before committing to this design (9/9 showcase pass, SHA-256 receipts).

**Wardley is ArcKit's own renderer.** Archify has no Wardley diagram type, and a Wardley Map's continuous evolution/value-chain plane — where the coordinates carry the meaning — does not fit its discrete node-graph IR. The real gap was that `/arckit:wardley` ended at a code block you had to paste into `create.wardleymaps.ai`, which sends the map off-estate and rules the flow out for OFFICIAL-SENSITIVE and air-gapped work.

## Changes

**New scripts**

- `owm-parse.mjs` — the single OWM reading of record, shared by the Mermaid and HTML paths so they cannot drift apart. Components, anchors, links, pipelines (both grammars), evolve targets, annotations, notes, sourcing directives, inertia.
- `owm-to-html.mjs` — compiles OWM to a self-contained interactive map: banded evolution axis, value chain, sourcing rings counted in a legend, evolve arrows, inertia markers, pipelines, annotations with leader lines. ~26KB for an 11-component map; embeds its OWM source so maps round-trip.
- `archify-detect.mjs` — locates Archify across `$ARCKIT_ARCHIFY_HOME`, project-local and per-runtime global skill directories. Reports version, CLI path and available types.

**Commands** — new `archify.md`; `wardley.md` gains the **Interactive HTML Map** step (on by default, `--no-html` to skip).

**Docs** — `guides/archify.md`, both CHANGELOGs, DEPENDENCY-MATRIX, `llms.txt`, command counts 75 → 76, site indices.

## Defects found and fixed in review

A self-review of the branch found six defects, every one failing **silently**, four of which broke `/arckit:wardley`'s own documented worked example. All were reproduced before fixing (`fc4732c`):

- **Shipping** — the three new scripts were missing from `converter.py`'s `core_only_copies`, so no generated extension received them: `/arckit:wardley` broke mid-run and `/arckit:archify` was entirely non-functional on all seven non-Claude runtimes, with CI green throughout.
- **XSS** — `focusNode` concatenated a component name into `detail.innerHTML`, so an `img`/`onerror` payload executed on click. Reproduced in Chromium; the panel is now built with `createElement`/`textContent` and no `innerHTML` sink remains.
- **Parser** — pipeline block children carrying only an evolution coordinate were dropped; the `annotation 1,[…]` comma form was rejected outright; multi-point annotations lost points and leaked coordinates into the text; and the directive skip-list ran before the link rule, so `Title Service -> API` was parsed as a `title` statement, silently renaming the map.

The parser now also reports unknown link endpoints by end, and reports lines no rule claims, instead of discarding them.

## Verification

- **1888 pytest, 407 node tests**, all 17 lint-workflow gates and markdownlint clean.
- Rendering checked in Chromium at 1600×1000 in both themes: no JS errors, no horizontal overflow, dependency tracing correct.
- `tests/plugin/test_command_scripts_ship.py` derives the shipping check from the commands themselves and was verified to fail against the pre-fix converter.
- 42 renderer/parser cases and 9 detection cases, including one that parses the command's own worked example and asserts zero warnings.

https://claude.ai/code/session_01DkfB3Gj4v4SrnLKvkPfJFt


johnfelipe:
how can test this in my own clone repo, but im using git bash for run commands, may be is possible you create .sh script for run and see all archified?

tractorjuice:
@johnfelipe this is merged now and will ship in the next release. You don't need a `.sh` script. Both helpers are plain Node scripts, so they run the same in Git Bash as on Linux or macOS. Node 18 or later is all you need.

**Inside Claude Code** (once the release is out and your plugin has updated):

```
/arckit:archify wardley 001
/arckit:archify architecture booking platform
```

**From your own clone in Git Bash, without Claude Code:**

1. Wardley Maps work straight away because the renderer is built in. Copy the OWM code block out of one of your `ARC-*-WARD-*.md` maps into a file, then:

   ```bash
   node plugins/arckit-claude/scripts/owm-to-html.mjs map.owm map.html --title "My Map"
   start map.html   # opens it in your browser from Git Bash
   ```

   The HTML is a single self-contained file that makes no network requests.

2. The other types (`architecture`, `workflow`, `sequence`, `dataflow`, `lifecycle`) need the third-party [Archify](https://github.com/tt-a1i/archify) skill installed first:

   ```bash
   npx skills add tt-a1i/archify -g
   node plugins/arckit-claude/scripts/archify-detect.mjs --json   # should report "found": true
   ```

   If it isn't found, set `ARCKIT_ARCHIFY_HOME` to where you installed it. Rendering these types happens inside `/arckit:archify`, because the model writes Archify's diagram IR from your artefacts. There isn't a standalone "archify everything" script.

`docs/guides/archify.md` has the full details. If anything breaks on Windows, please open an issue with the command and its output.

